PR #2, opened 2019-03-20
Simplify dataset reader

dcfidalgo:
This is a first step to simplify the dataset readers.
It also updates the tests a bit.

- the idea is to tie them to the models -> they are named accordingly
- the names of the forward parameters are directly taken from the model's forward method
- the label field is identified by checking if it is optional in the model's forward method 

Follow-up MR will be:
- Update the ExamplePreparator
- Update tests properly

All tests pass locally for me.
@dvsrepo @frascuchon can you have a quick look at it? i would like to merge this soon, to update the ExamplePreparator next.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,5 @@ tools
 generated*
 
 *venv
+
+.DS_Store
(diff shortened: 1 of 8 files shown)


dvsrepo:
No comments, looks great!

I like the use of signature.

PR #18, opened 2019-04-11
feat: allow datasource paths relatives to datasource yaml definition

frascuchon:


dcfidalgo:
@frascuchon I merged #19 since i needed it for evo. I will leave this PR open, for future discussion when we implement the new data source cfg format.
Also i like the idea of switching to classes for the reader methods (`from_*` methods), but we have to think about how to combine this abstract class thing with `biome-core`.

dvsrepo:
@dcfidalgo is this still relevant or can we close it?

dcfidalgo:
@dvsrepo can be closed. i will close it.

PR #19, opened 2019-04-30
Allow relative paths in the data source cfg yaml file

dcfidalgo:


dcfidalgo:
I need this for evo, will merge it now as is.
We can reopen the discussion when we implement the new data source cfg scheme.

PR #59, opened 2019-10-03
feat: avoid create allennlp fields without value

frascuchon:


dcfidalgo:
This changes quite a lot the behavior described in [this pitch](https://gitlab.com/recognai-team/team/issues/31). I will copy the content of the pitch to an issue here in github, and change the title to something like: "How AllenNLP deals with empty Fields".

With this PR, if you have an empty input, but no empty label, the training will throw a KeyError, since a required input is missing in the Instance:
```    
    instance_with_lengths = ([padding_lengths[field_name][padding_key]
>                             for (field_name, padding_key) in sorting_keys],
                             instance)
E   KeyError: 'tokens'
```
If we have an empty label, but non-empty input, it will throw an error depending on the batch size:
- if it is the only Instance in the batch with a missing LabelField:
```
allennlp.common.checks.ConfigurationError: 'You cannot construct a Batch with non-homogeneous Instances.'
```
- If all Instances in the batch are missing the LabelField -> `RuntimeError: 
```
RuntimeError: The model you are trying to optimize does not contain a 'loss' key in the output of model.forward(inputs).
```

For me it's ok to merge this now, since it makes sure that it will always fail when there is an empty string in either the input or the label. But it will not fail if there are empty strings in the input AND the label.

In general, i would prefer a more "explicit" way of how we deal with empty tokens, with meaningful error messages. Maybe some function/class, with the only purpose to make some explicit modifications to the input data, before starting the whole allennlp field/instance thing.
@frascuchon @dvsrepo open for discussion!

PR #62, opened 2019-10-08
Adapt the `SimilarityClassifier`

dcfidalgo:
This PR makes the `SimilarityClassifier` inherit from the `BaseModelClassifier`.
This means that it now works with `ListField`s as well.

The only difference to the `SequencePairClassifier` model, is that the `SimilarityClassifier` adds another term to the loss which is a function of the cosine similarity of the two records.
I tried to unify the `SimilarityClassifier` and the `SequencePairClassifier`, but i saw that there would be too many changes necessary for the `SequencePairClassifier` model, so i left the two models separated.

@frascuchon @dvsrepo I also added an explicit exception if a multifield_seq2seq encoder is added in the `BaseModelClassifier`, since there is a TODO, that this is not working ... Is this still true?

I tested the two models with my synthetic language data set, the `SimilarityClassifier` performs a bit better, but only by a small margin:
Screenshot 2019-10-08 at 13 41 14

### DIFF
--- a/src/biome/text/models/base_model_classifier.py
+++ b/src/biome/text/models/base_model_classifier.py
@@ -20,7 +20,7 @@ from torch.nn import Dropout, Linear
 
 
 class BaseModelClassifier(Model, metaclass=ABCMeta):
-    """This ``BaseModelClassifier`` simply encodes a sequence with a ``Seq2VecEncoder``, then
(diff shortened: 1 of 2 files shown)


dvsrepo:
This is great!

Yes, multifield_seq2seq is still not working so the exception is correct.

PR #63, opened 2019-10-09
Feature/concrete text to instances parameters

frascuchon:
This PR changes de general method param name for the `text_to_instance(text: dict)` methods  and apply the concrete transformations for generate a properly `Instance` example


src/biome/text/dataset_readers/sequence_classifier_dataset_reader.py
```diff
@@ -103,60 +85,37 @@ def _read(self, file_path: str) -> Iterable[Instance]:
         return (instance for idx, instance in instances.iteritems() if instance)
 
     def text_to_instance(
-        self, text: Union[Dict, "dask.Series", "pandas.Series"]
+        self,
+        tokens: Union[str, List[str], dict],
+        label: Optional[str] = None,
+        # TODO we can use the extra args for metadata build. Not used for now but required
+        **extra_args
```
dcfidalgo: For now i would just remove this argument, and introduce it when we have a concrete example for the metadata ... YAGNI ... :)

frascuchon: If we remove the keyword variables, the training process will fail since we're passing all the values read from `DataSource` 
```python
                lambda d: self.text_to_instance(
                    **{str(k): v for k, v in d.to_dict().items()}
                ),
                axis=1,
                meta=(None, "object")
```

We need a mechanism for filter just the `text_to_instance` to do that. I'll think about it and push the changes

src/biome/text/dataset_readers/sequence_classifier_dataset_reader.py
```diff
@@ -103,60 +85,37 @@ def _read(self, file_path: str) -> Iterable[Instance]:
...
+        if tokens_field:
+            fields["tokens"] = tokens_field
+        if label_field:
+            fields["label"] = label_field
 
-        if param_name == self.LABEL_TAG:
-            return LabelField(value)
-        else:
-            return self.build_textfield(value)
+        return Instance(fields) if fields and len(fields) > 0 else None
```
dcfidalgo: I think this is a duplicated check...
a non-empty dict will always evaluate to true and the length of an non-empty dict is always greater than 0.

src/biome/text/dataset_readers/sequence_classifier_dataset_reader.py
```diff
@@ -103,60 +85,37 @@ def _read(self, file_path: str) -> Iterable[Instance]:
         return (instance for idx, instance in instances.iteritems() if instance)
 
     def text_to_instance(
-        self, text: Union[Dict, "dask.Series", "pandas.Series"]
+        self,
+        tokens: Union[str, List[str], dict],
```
dcfidalgo: I think the `build_textfield` method does not accept a `List[str]`

src/biome/text/dataset_readers/sequence_classifier_dataset_reader.py
dcfidalgo: ah ok, i did not understand this part. I think the solution you have is fine with these extra_kwargs that are thrown away basically, it should just be better documented in the doc string :)

dcfidalgo: or a better name for this parameter, like `trash_kwargs` or something like this

src/biome/text/dataset_readers/sequence_classifier_dataset_reader.py
frascuchon: I've made minimal changes for that

src/biome/text/dataset_readers/sequence_classifier_dataset_reader.py
frascuchon: Given the base class for DatasetReader, I have made some changes to avoid having to define that parameter **extra_args

PR #70, opened 2019-10-22
Fix/adapt to new datasource format

dcfidalgo:
This PR is related to https://github.com/recognai/biome-data/pull/9

It simply adapts biome-text to the new data source format.

I also removed the `.split()` action from the label in the `text_to_instance` methods, since for strings with only space characters it was buggy.

I will provide a new PR later today, in which we deal with empty tokens/labels in a way we discussed [here](https://gitlab.com/recognai-team/team/issues/31). Right now, empty tokens/labels will crash/mess up the training process.


dcfidalgo:
The tests fail, because we first need to merge the PR in biome-data ...

PR #72, opened 2019-10-24
define a environment module for all env vars definitions

frascuchon:
A simple PR for refactoring the use of environment variables by defining a new python module where all environment variables used in project should be defined

### DIFF
--- a/src/biome/text/commands/explore/explore.py
+++ b/src/biome/text/commands/explore/explore.py
@@ -7,7 +7,7 @@ import re
 from allennlp.commands.subcommand import Subcommand
 from allennlp.common.checks import ConfigurationError
 from biome.data.sources import DataSource
-from biome.data.utils import ENV_ES_HOSTS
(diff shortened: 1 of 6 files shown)


src/biome/text/commands/explore/explore.py
```diff
@@ -7,7 +7,7 @@
 from allennlp.commands.subcommand import Subcommand
 from allennlp.common.checks import ConfigurationError
 from biome.data.sources import DataSource
-from biome.data.utils import ENV_ES_HOSTS
+from ...environment import ES_HOST, BIOME_EXPLORE_ENDPOINT
```
dcfidalgo: I think in general absolute imports are preferred over relative ones except for a few cases (https://realpython.com/absolute-vs-relative-python-imports/)

src/biome/text/commands/helpers.py
```diff
@@ -5,6 +5,8 @@
 
 import yaml
 
+from ..environment import CUDA_DEVICE
```
dcfidalgo: Same comment as above

PR #73, opened 2019-10-24
Fix show label on ui

frascuchon:
This PR solves problems with biome-ui showing label value as part of input data.

see 
Captura de pantalla 2019-10-24 a las 13 23 13

The reason is that, using general signature as input information, the label field are included. We define a new predict_signature that will drop the optional parameters (and no required for inference)


frascuchon:
@dcfidalgo  totally agree with you.

I've made some changes for manage that information internally to function `register_biome_prediction`

PR #74, opened 2019-10-25
Feature/seperate df from mapped df

dcfidalgo:
This adopts biome-text to the new `to_mapped_dataframe` method introduced in [this PR](https://github.com/recognai/biome-data/pull/13).

It also applies a str transformation to the labels and skips examples when label is an empty string. The implementation is awful and i will improve this in a separate PR soon.
The prediction also returns a None now, if the example could not be transformed to an instance.

It also makes the pipeline tests faster by simplifying the model and trainer.

### DIFF
--- a/src/biome/text/commands/explore/explore.py
+++ b/src/biome/text/commands/explore/explore.py
@@ -4,10 +4,13 @@ import os
 import argparse
 import datetime
 import re
+import warnings
+
 from allennlp.commands.subcommand import Subcommand
 from allennlp.common.checks import ConfigurationError
 from biome.data.sources import DataSource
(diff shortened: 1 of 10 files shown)


dcfidalgo:
@frascuchon the fails are related to the latest PR in biome-data i think (Using dask elk for datasource)

frascuchon:
> @frascuchon the fails are related to the latest PR in biome-data i think (Using dask elk for datasource)

I think there was a problem resolving conflicts. This changes where merged into master in this PR

See this https://github.com/recognai/biome-text/commit/d505143c3fd9b8c1094f77d45d525ca252db2c3e

PR #78, opened 2019-11-04
Define training callbacks

frascuchon:
This PR includes some works related to [Pitch#27](https://gitlab.com/recognai-team/team/issues/27)

The idea is extends default callback trainer functionalities for 2 purposes:

* Customize and controls the log messages
* Evaluate the test dataset after training (This is controlled by evaluate_on_test param in default trainer, but the callback trainer doesn't support it)

There is a callback trainer configuration example [here](https://gitlab.com/recognai-team/biome/biome-classifier-workflows/blob/master/workflows/tests/classifier.char.list.field/trainer/callback.trainer.yml) that you can check and test

```yaml
iterator:
  batch_size: 64
  cache_instances: 'true'
  max_instances_in_memory: 200000
  sorting_keys:
  - - tokens
    - num_fields
  type: bucket
trainer:
  type: callback
  num_epochs: 3
  # patience: 1
  optimizer:
    amsgrad: 'true'
    type: adam
  # validation_metric: +accuracy
  callbacks:
  - type: gradient_norm_and_clip
    grad_norm: 1.0
  - checkpoint
  - type: track_metrics
    patience: 500
  - validate
  - type: log_to_tensorboard
    log_batch_size_period: 10
  - evaluate # This is our EvaluateCallback
  - logging  # This is our LoggingCallback
```

This is a generic allennlp callback trainer configuration. But the idea in this PR is define a mechanism for automatic create and configure the callback trainer que different callbacks using the original trainer configuration


src/biome/text/commands/helpers.py
```diff
@@ -145,7 +146,9 @@ def to_allennlp_params(self) -> Dict:
         for field, path in zip(data_fields, data_paths):
             if path:
                 allennlp_params[field] = path
-        if self.test_path:
-            allennlp_params[self.EVALUATE_ON_TEST_FIELD] = True
 
+        trainer_cfg = allennlp_params[self.TRAINER_FIELD]
+        if not trainer_cfg.get("type"):  # Just in case of default trainer
+            trainer_cfg["type"] = DefaultCallbackTrainer.__name__
+            allennlp_params[self.EVALUATE_ON_TEST_FIELD] = False
```
dcfidalgo: Does this mean that if i want to use the `default` trainer (no callback trainer) i have to write:
```yaml
trainer:
    type: default
```
in my trainer yaml config?

frascuchon: Exactly. Just in case of no trainer type is defined, the callback trainer is used as "default". If you set any type, it will be used as the trainer.

But, the idea is to be 100% compatible with the default ones configuration. So, have you a configuration sample of old configuration that doesn't work with the new trainer ?

dcfidalgo: I had some trainer.yml files where i specified the `summary_interval`. I added this parameter to the `DefaultCallbackTrainer` and pushed the changes. If i encounter more cases i will either add them as well, or add this `type: default` line to the config file.

PR #82, opened 2019-12-11
Implement AllenNLP's BiMPM model in biome-text

dcfidalgo:
This PR implements the BiMPM model in biome-text as `"biome_bimpm"`.

For the implementation i created a mixin class `BiomeClassifierMixin` that allows its child classes to work with Biome's UIs by generalizing the `decode` and `get_metrics` methods. I added an example of how to implement AllenNLP models to work with Biome's UIs.

@frascuchon Given that we have this mixin class now, i am not sure if the `BaseModelClassifier` adds much value. Maybe it would be cleaner to remove it and move its model architecture logic to the `SequenceClassifier`. The `SequencePairClassifier` can then directly inherit from the `SequenceClassifier` and call its forward method twice to accomplish the siamese network. What do you think?

### DIFF
--- a/src/biome/text/dataset_readers/sequence_pair_classifier_dataset_reader.py
+++ b/src/biome/text/dataset_readers/sequence_pair_classifier_dataset_reader.py
@@ -48,3 +48,4 @@ class SequencePairClassifierReader(DataSourceReader):
 
 # Register an alias for this reader
 DatasetReader.register("similarity_classifier")(SequencePairClassifierReader)
+DatasetReader.register("biome_bimpm")(SequencePairClassifierReader)
--- a/src/biome/text/dataset_readers/sequence_pair_classifier_dataset_reader.py
+++ b/src/biome/text/dataset_readers/sequence_pair_classifier_dataset_reader.py
@@ -48,4 +48,3 @@ class SequencePairClassifierReader(DataSourceReader):
 
 # Register an alias for this reader
 DatasetReader.register("similarity_classifier")(SequencePairClassifierReader)
-DatasetReader.register("biome_bimpm")(SequencePairClassifierReader)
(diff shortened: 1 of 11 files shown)


dcfidalgo:
Just realized that i forgot to implement a corresponding pipeline ...

src/biome/text/dataset_readers/sequence_pair_classifier_dataset_reader.py
```diff
@@ -48,3 +48,4 @@ def text_to_instance(
 
 # Register an alias for this reader
 DatasetReader.register("similarity_classifier")(SequencePairClassifierReader)
+DatasetReader.register("biome_bimpm")(SequencePairClassifierReader)
```
frascuchon: It this line an error? Do we need define this line? It's not managed by corresponding biome pipeline class?

src/biome/text/models/base_model_classifier.py
```diff
@@ -19,7 +18,141 @@
...
+
+        num_metrics = len(self._biome_classifier_metrics)
+        all_metrics["average/f1"] = total_f1 / num_metrics
+        all_metrics["average/precision"] = total_precision / num_metrics
+        all_metrics["average/recall"] = total_recall / num_metrics
+
+        return all_metrics
+
+
+class BaseModelClassifier(BiomeClassifierMixin, Model):
```
frascuchon: Now, the BaseModelClassifier class is not a proper base class, but keep shared code for our SequenceClassifier and SequencePairClassifier implementations for the architecture initialization. The only value is changing is the number of input. 1 for sequence classifier and 2 for sequence pair classifier. 

It should be nice redefine those architectures for manage with one single class, avoiding this messy heritage 

Anyway, the naming used is not the best right now. We can call it `SequenceClassifierBase` or `BaseSequenceClassifier`, while we think how to consolidate sequence and sequence-pair in a more elegant way

dcfidalgo:
I implemented the comments by Paco (moved register lines to the pipeline class file, we quickly talked about it + chose a better naming for the base class for the sequence classifiers).

> It should be nice redefine those architectures for manage with one single class, avoiding this messy heritage

In this case i would not mind making the `SequencePairClassifier` a child of the `SequenceClassifier`, since it really just inherits the model architecture and basically applies it two times ... and then just get rid of the `SequenceClassifierBase`

PR #88, opened 2020-01-13
Multifield bimpm model

dcfidalgo:
This PR introduces a multifield BiMpm model. I tested it with my language data set and it seems to work quite well, i will add the experiment to the `biome-classifiers` repo as soon as possible.

Since this model is very specific to DZ Bank / Uniserv data (several records per golden record), i am not entirely sure if we want to include it in the library. We could also move it to our private DZ Bank repo and then load it via the `--include-package` argument.
@frascuchon @dvsrepo what do you think?


frascuchon:
It makes sense for me publish as part of open source library

dvsrepo:
Same for me.

PR #89, opened 2020-01-14
feat: add datasource and model as part of explore attributes

frascuchon:
This PR just includes datasource and model as part of explore action attributes information.

It could be nice calculate basic model metrics  (F1, recall and precision) using some `dask.dataframe` directives and add them too.

@dcfidalgo @dvsrepo Should be possible?


dcfidalgo:
> @dcfidalgo @dvsrepo Should be possible?

Sip, i think it shouldn't be too difficult.

PR #97, opened 2020-02-10
feat(reader): transform lists to strings for text fields

frascuchon:
This PR converts incoming list into string values properly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ default: help
 
 check: ## applies a code pylint with autopep8 reformating
 	@black .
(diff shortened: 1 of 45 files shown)


frascuchon:
> 👍 I pushed a commit in which we use the new method `_value_to_string` more efficiently. Before we stringified the input two times, once outside the method and then again inside the method.
> 
> If you prefer it the other way, just omit my commit and merge as is!

It's ok for me. Just try to fix the failed tests, please

dcfidalgo:
Should be fixed now

PR #103, opened 2020-02-18
switch to python3 namespace package

dcfidalgo:
@frascuchon This is the change we talked about yesterday.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 import os
 
 from pip import __version__ as pip_version
-from setuptools import setup, find_packages
+from setuptools import setup, find_namespace_packages
 from typing import Tuple
 
 PIP_VERSION_REQUIRED = "19.1.1"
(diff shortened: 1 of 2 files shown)


frascuchon:
Just to be sure. We need all namespaced modules configured in this way if we install from a package distribution wheel??

I mean, if we merge this change and without change the rest of the packages (biome-data, biome-classifier, extraction...) Will we break the package ecosystem installation?

dcfidalgo:
No, following [these docs](https://packaging.python.org/guides/packaging-namespace-packages/) the two ways are compatible:
- native python 3 namespace packages, that is without any init file in the namespace package
- pkgutil-style namespace packages, that is an init file that contains **only** 
`__path__ = __import__('pkgutil').extend_path(__path__, __name__)`

What will break, is the autocomplete function for Pycharm, somehow it can not handle both cases at the same time.

dcfidalgo:
@frascuchon can we merge this PR and the similar ones in biome-data and biome-classifier?

frascuchon:
> @frascuchon can we merge this PR and the similar ones in biome-data and biome-classifier?

We will start merging biome-data and see the changes in dev. If everything if fine, then we can start merge the other ones

PR #108, opened 2020-03-03
save pipeline predictions to file

### LINKED ISSUE #107: Allow trace requested predictions for retraining purposes
Enable trace mode in pipeline for keep all input/predictions results for a given serving model.

frascuchon:
This should close #107

Need to include:

- parameterizable enable/disable behaviour
- parameterizable output path 
- add tests and checks
- check elasticsearch ingestions are ready for annotate


dcfidalgo:
@frascuchon I am still missing the check about the compatibility with ES. The rest should be fine.

frascuchon:
> @frascuchon I am still missing the check about the compatibility with ES. The rest should be fine.

It's ok for me. 

We must include keep prediction on batch prediction but we will cover in another PR

Thanks!!

PR #111, opened 2020-03-04
add possibility to cache predictions

dcfidalgo:
This PR adds the possibility to cache the predictions.

I did not manage to make the cache size configurable in a reasonable way, so i just added a switch to the explore command: either you want to cache the predictions or not.


src/biome/text/commands/explore/explore.py
```diff
@@ -137,6 +146,10 @@ def explore(
             "\nwith the same name that your model."
         )
 
+    predict = pipeline.predict_json
+    if cache_predictions:
+        predict = pipeline.predict_json_with_cache
```
frascuchon: I prefer manage cache activation/deactivation  inside "official" predictor method (`predict_json`) instead create another api method

src/biome/text/pipelines/pipeline.py
```diff
@@ -272,6 +273,33 @@ def predict_json(self, inputs: JsonDict) -> Optional[JsonDict]:
...
+        inputs
+            The input features/tokens in form of a json dict
+
+        Returns
+        -------
+        output
+            The model's prediction in form of a dict.
+            Returns None if the input could not be transformed to an instance.
+        """
+        pickled_inputs = pickle.dumps(inputs)
```
frascuchon: Could we create a hashable dict available for cache parameter hashing? I have the impression that it is much more efficient than pickle/unpickle the dictionary 

```python
class Hashabledict(dict):
    def __hash__(self):
        return hash(frozenset(self))

hashable_inputs = Hashabledict(inputs)
predict(hashable_inputs)
```

src/biome/text/pipelines/pipeline.py
```diff
@@ -272,6 +273,33 @@ def predict_json(self, inputs: JsonDict) -> Optional[JsonDict]:
...
+
+        return self.predict_json_pickle(pickled_inputs)
+
+    @lru_cache(maxsize=None)
+    def predict_json_pickle(self, inputs: bytes):
```
frascuchon: In any case,  `predict_json_pickle` should not be a public method

src/biome/text/pipelines/pipeline.py
```diff
@@ -272,6 +273,33 @@ def predict_json(self, inputs: JsonDict) -> Optional[JsonDict]:
...
+        -------
+        output
+            The model's prediction in form of a dict.
+            Returns None if the input could not be transformed to an instance.
+        """
+        pickled_inputs = pickle.dumps(inputs)
+
+        return self.predict_json_pickle(pickled_inputs)
+
+    @lru_cache(maxsize=None)
```
frascuchon: We must define a maxsize limit  for avoid memory overflow

dcfidalgo:
Addressed all comment by @frascuchon and made the prediction cache size configurable.
By default there is no cache initialized. You have to call the `Pipeline.init_prediction_cache(max_size)` method to init the cache.

Issue #112, opened 2020-03-06
Text Transform component


frascuchon:
Define a common clean pipeline that converts html content into extracted textual data.

dcfidalgo:
After talking to @frascuchon the idea is to create a new "DataCleaner" class (name is still to be decided) that inherits from AllenNLP's `Registrable`.
We can then add custom classes of this DataCleaner in the config file that implement certain cleaning methods that will be applied.

dvsrepo:
Discussed with David to:

1. Take into account the design of other libraries:
   - https://pytorch.org/docs/stable/torchvision/transforms.html
   - https://www.tensorflow.org/tfx/transform/get_started
   - https://dev.fast.ai/text.core#Preprocessing-rules (fastai v2)
  - https://docs.fast.ai/text.transform.html (fastai v1)

2. Allow extensibility by letting developers define their own Transform in their own modules (without using --include-package if possible)

PR #113, opened 2020-03-14
Feat/text transforms

dcfidalgo:
This PR addresses #112 

For this PR i first did a small refactoring by removing the `TextBuilderMixin` out of two reasons:
- Only one class was using this mixin
- A lot of parameters had to be passed through the `DataSourceReader` to the `TextBuilderMixin`, which in general increases the complexity of the code.

Then we implement a `TextTransforms` class that performs some transformation to a `str` by simply calling one of its instances. For example:
```python
from biome.text.dataset_readers.text_transforms import Html2TextTransforms
transform = Html2TextTransforms()
transform("<p> test this    amp;   that shit!  </p>")
# "test this & that shit!"
```

There are two ways this can be customized through the model yaml config:
 1. You can provide a list of rules supposing these rules are already implemented in the class as class methods:
```yaml
pipeline:
    tokenizer:
        type: word
    token_indexers:
        tokens:
            type: single_id
            lowercase_tokens: true
    text_transforms:
        rules:
        - "strip_spaces"
```
 2. You can provide your own `TextTransforms` implementation:
```yaml
pipeline:
    tokenizer:
        type: word
    token_indexers:
        tokens:
            type: single_id
            lowercase_tokens: true
    text_transforms:
        type: "html_to_text"
        # or for a custom implementation outside of biome-text (though i haven't tried this ...)
        # type: mymodule.MyTextTransforms
```

Both ways can be combined. The default setting for now would be:
```yaml
    text_transforms:
        type: "rm_spaces"
        rules:
        - "strip_spaces"
        - "rm_useless_spaces"
```

@dvsrepo @frascuchon comments, suggestions?


dcfidalgo:
I implemented a `html_to_text` transform using the BeautifulSoup4 package and the `lxml` parser (in included them in the *setup.py*).
Also added some tests.

I am not sure about the `fix_html` method in the `Html2TextTransforms`, i know close to nothing about html and i just copied it from fastai. Do you think this method is necessary @frascuchon ?

frascuchon:
> I implemented a `html_to_text` transform using the BeautifulSoup4 package and the `lxml` parser (in included them in the _setup.py_).
> Also added some tests.
> 
> I am not sure about the `fix_html` method in the `Html2TextTransforms`, i know close to nothing about html and i just copied it from fastai. Do you think this method is necessary @frascuchon ?

It's depends on model behaviour. `BeautifulSoup.getText` method transforrms some especial characters into unicode representation (`nbsp;` --> `\xa0`). If the unicode representation is ok for model training/prediction, we can remove it, otherwise, the pipeline always should be applied  as 
`fix_html` -> `soup.getText`.

dvsrepo:
> > I implemented a `html_to_text` transform using the BeautifulSoup4 package and the `lxml` parser (in included them in the _setup.py_).
> > Also added some tests.
> > I am not sure about the `fix_html` method in the `Html2TextTransforms`, i know close to nothing about html and i just copied it from fastai. Do you think this method is necessary @frascuchon ?
> 
> It's depends on model behaviour. `BeautifulSoup.getText` method transforrms some especial characters into unicode representation (`nbsp;` --> `\xa0`). If the unicode representation is ok for model training/prediction, we can remove it, otherwise, the pipeline always should be applied as
> `fix_html` -> `soup.getText`.
> 
> @dvsrepo what do you think?

The unicode representation is not useful for the model. 

In order to asses if we include fix_html -> soup.getText we should make sure they play nicely together (fix_html does replacements of html-specific things and it's not guaranteed to output HTML which can be parsed with BS).

@dcfidalgo have you tested the fix_html --> html_to_text pipeline?

dcfidalgo:
Ok, yesterday i read all about unicode/character encoding/python strings and i think i understand it now more or less ... [This](https://realpython.com/python-encodings-guide/) was quite helpful.

BeautifulSoup stores everything (also the HTML entities) as unicode code points, the same as python3 str. Now, when you display a str variable in python, it tries to display the unicode code points as human readable chars, which in the case of the `0xa0` code point is not possible, since it is a "non-breakable space" char (actually when you do `print("\xa0")` it will just display a white space, the same happens when you write it to a file `with open("test.txt", "w") as file: file.write("\xa0") ). I am not sure what the result is when building an AllenNLP vocab with this unicode code point, maybe it just works ...

Anyways, for us i do not think there is a real difference between a blank space and a non-breaking space, so the updated `fix_html` method just replaces a nbsp with a blank space. I also looked for other "non-human readable chars" in the html entities, but could not find any, so i think this is maybe the only rule we need.

In addition i also added a replace rule for the `<br>` html single break symbol, which now gets replaced with a simple `\n`.

If we find issues in the future we can just add those replacements to the `fix_html` method, but i think in general we should be fine like this.

@frascuchon @dvsrepo if you agree, i would merge this PR this afternoon.

PR #114, opened 2020-03-19
Feature/simplify pipeline definition with generics

frascuchon:
This PR simplifies pipeline definitions by using class generics. So, for create a new pipeline class definition, you don't need implement `model_class` and `reader_class` methods:

```python
class MySimplePipeline(Pipeline[AwesomeArchitecture, BetterReader]):
   # Eventuallly you need to implement specific predict method
   pass
```


src/biome/text/pipelines/similarity_classifier.py
```diff
@@ -1,30 +1,18 @@
-from typing import Union, List, Type
+from typing import Union, List
 
-from allennlp.models import Model as AllennlpModel
 from allennlp.predictors import Predictor
 
-from biome.text.pipelines.pipeline import Pipeline
 from biome.text.dataset_readers import SequencePairClassifierReader
-from biome.text.dataset_readers.datasource_reader import DataSourceReader
 from biome.text.models import SimilarityClassifier as TorchSimilarityClassifier
```
dcfidalgo: I think there is no more need for this renaming
```suggestion
from biome.text.models import SimilarityClassifier
```
I think this was done to not clash with the pipeline class name, but now it's `SimilarityClassifierPipeline`

PR #116, opened 2020-03-24
Feat/entity classifier

dcfidalgo:
This PR introduces the `EntityClassifier` pipeline.

For now it just duplicates functionality, using the `SequenceClassifier` as model. Compared to using the `SequenceClassifierReader`, a data source yaml for the `EntityClassifierReader` could look like this:
```yaml
source: myfile.csv
mapping:
    kind: kind
    context: context
    position: position
    column_header: column_header
    label: label
```

@frascuchon I treated the `end char position` as exclusive, please tell me if this is correct.

@frascuchon @dvsrepo Still to discuss: 
* How to avoid identity mappings (see yaml above)
* I am pretty sure the reader will change substantially once we have a dedicated model for the `EntityClassifer`

### DIFF
--- a/src/biome/text/dataset_readers/entity_classifier_reader.py
+++ b/src/biome/text/dataset_readers/entity_classifier_reader.py
@@ -1,0 +1,78 @@
+from typing import Union, List, Optional, Tuple
+
+from allennlp.data import Instance
+from biome.text.dataset_readers.datasource_reader import DataSourceReader
+
+
+class EntityClassifierReader(DataSourceReader):
+    def text_to_instance(
+        self,
+        kind: str,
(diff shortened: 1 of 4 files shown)


frascuchon:
@dcfidalgo 

Could we rewrite this PR in terms of new api desing?

dcfidalgo:
@frascuchon Not sure, this PR is mainly an implementation of a DatasetReader, but in the new design we just keep a generic one, right? Let's discuss it in a quick call ...

dcfidalgo:
I will close this PR for now. A new one will come after revising the use case and taking into account the new library design.

PR #119, opened 2020-03-25
feat: package redefiniton

frascuchon:
## Introduction

This PR creates a new pipeline design based on internal discussions.

A lot of changes will be found here, so, a review tracking will be hard to do. This is the reason why every commit comes from a different pull request where you can check (too late for review) the partial changes

## Keep in mind

- The old pipeline implementations are fully operatives
- The command line works using the old implementation/configuration
- You can use these new features using as python library

## An example of use
See [examples](https://github.com/recognai/biome-text/tree/feature/hierarchical-pipeline-definition/examples) folder


dvsrepo:
Hi!

I started testing this yesterday and run into the following error when running the example `examples/text_classification.py`:

```
Traceback (most recent call last):
  File "examples/text_classification.py", line 43, in <module>
    """
  File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/biome/text/defs.py", line 173, in from_config
    return pipeline_class(**config)
  File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/biome/text/pipelines/_impl/allennlp/allennlp_pipeline.py", line 126, in __init__
    self.init_extra_configuration(**extra_configuration)
  File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/biome/text/pipelines/_impl/allennlp/text_classifier.py", line 14, in init_extra_configuration
    self._wrapping_dims = 0 if self._aggregate_inputs else 1
  File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/torch/nn/modules/module.py", line 576, in __getattr__
    type(self).__name__, name))
AttributeError: 'TextClassifierPipeline' object has no attribute '_aggregate_inputs'
```
I will take look later

dvsrepo:
I have commited a suggested change for solving the above issue.

Now I am running into the following error, connected to the forward and text to instance functions :

```
Traceback (most recent call last):
  File "/Users/dani/recognai/testing/biome-text/examples/text_classification.py", line 49, in <module>
    validation="text_classifier/validation.data.yml",
  File "/Users/dani/recognai/testing/biome-text/src/biome/text/pipelines/_impl/allennlp/allennlp_pipeline.py", line 317, in learn
    _ = learn(model_spec=spec, **kwargs)
  File "/Users/dani/recognai/testing/biome-text/src/biome/text/pipelines/_impl/allennlp/learn/learn.py", line 98, in learn
    recover=is_recovered,
  File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/allennlp/commands/train.py", line 252, in train_model
    metrics = trainer.train()
  File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/allennlp/training/trainer.py", line 478, in train
    train_metrics = self._train_epoch(epoch)
  File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/allennlp/training/trainer.py", line 320, in _train_epoch
    loss = self.batch_loss(batch_group, for_training=True)
  File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/allennlp/training/trainer.py", line 261, in batch_loss
    output_dict = self.model(**batch)
  File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/torch/nn/modules/module.py", line 532, in __call__
    result = self.forward(*input, **kwargs)
TypeError: forward() got an unexpected keyword argument 'tokens'
```

frascuchon:
> I have commited a suggested change for solving the above issue.
> 
> Now I am running into the following error, connected to the forward and text to instance functions :
> 
> ```
> Traceback (most recent call last):
>   File "/Users/dani/recognai/testing/biome-text/examples/text_classification.py", line 49, in <module>
>     validation="text_classifier/validation.data.yml",
>   File "/Users/dani/recognai/testing/biome-text/src/biome/text/pipelines/_impl/allennlp/allennlp_pipeline.py", line 317, in learn
>     _ = learn(model_spec=spec, **kwargs)
>   File "/Users/dani/recognai/testing/biome-text/src/biome/text/pipelines/_impl/allennlp/learn/learn.py", line 98, in learn
>     recover=is_recovered,
>   File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/allennlp/commands/train.py", line 252, in train_model
>     metrics = trainer.train()
>   File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/allennlp/training/trainer.py", line 478, in train
>     train_metrics = self._train_epoch(epoch)
>   File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/allennlp/training/trainer.py", line 320, in _train_epoch
>     loss = self.batch_loss(batch_group, for_training=True)
>   File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/allennlp/training/trainer.py", line 261, in batch_loss
>     output_dict = self.model(**batch)
>   File "/Users/dani/recognai/testing/biome-text/.env/lib/python3.6/site-packages/torch/nn/modules/module.py", line 532, in __call__
>     result = self.forward(*input, **kwargs)
> TypeError: forward() got an unexpected keyword argument 'tokens'
> ```

The changes are ready to test. Yo can check the `examples` folder for different pipeline uses and implementations. 

The  new suggested module packages is under `biome.text.pipelines.api_new`

dvsrepo:
Hi!

Looks good. Just a few comments:

- the line separating `pipelines` class, `PipelineModel` in `base_model.py` and `TextClassifierBase` is a bit confusing. It is maybe about naming, but I think we should reduce to the minimum the number of concepts to handle, or have a very clear explanation about why each of this concepts. What is a Pipeline? What is a Model? What is a Task?

- `pipelines` class is intentionally lowercased? it is intended to have direct dependencies to AllenNLP? I would imagine we can have an abstract Pipeline with no deps to AllenNLP

- `vocabulary`  is intentionally lowercased? It does not extend AllenNLP Vocabulary? Could it be further abstracted?

It is a good chance to review naming and keep it as simple as possible for developers (e.g., we want to keep `learn` or go to more standard `train`).

dvsrepo:
I have added now a `TokenClassifier` example. Backbone model is (almost) equivalent to the TextClassification one, except from the num_wrapping_dim and TimeDistributedEncoder which do not make much sense outside hierarchical classification tasks.

It's not working yet due to CRF instantiation failing with num_labels=0 as we have no "outside vocab" when instantiating from_params. Will check this tomorrow morning.

dvsrepo:
@frascuchon see https://github.com/recognai/biome-text/blob/feature/hierarchical-pipeline-definition/examples/token_classifier.py

frascuchon:
lowercased classes are intended to manage the related classes (`pipelines` the pipeline model, and `vocabulary`  the `Vocabulary` classes) given an extra functionality that exceed the class bounds, like `entend_labels` for vocab, and `learn` or `explore` for pipeline models.

The `TextClassifierBase` define a bunch of method for the text classification specific task.  We  must bound this task since the ui applications depends task output formats and we cannot support arbitrary pipeline model output formats

frascuchon:
> I have added now a `TokenClassifier` example. Backbone model is (almost) equivalent to the TextClassification one, except from the num_wrapping_dim and TimeDistributedEncoder which do not make much sense outside hierarchical classification tasks.
> 
> Instantiation not working yet due to CRF instantiation failing with num_labels=0 as we have no "outside vocab" when instantiating from_params. Will check this tomorrow morning, next I will prepare example data and test the full pipeline.

Now it's working. There is a problem with output label key names (i think it's returning the tag label, but i'm not sure)

dvsrepo:
> lowercased classes are intended to manage the related classes (`pipelines` the pipeline model, and `vocabulary` the `Vocabulary` classes) given an extra functionality that exceed the class bounds, like `entend_labels` for vocab, and `learn` or `explore` for pipeline models.
> 
> The `TextClassifierBase` define a bunch of method for the text classification specific task. We must bound this task since the ui applications depends task output formats and we cannot support arbitrary pipeline model output formats

Ok, let's discuss this tomorrow, I think we can streamline this a lit a bit more: `PipelineModel` is still a strange concept, managed by `pipelines`, and under `models`, and should not implement methods like `text_to_instance` which is implemented and only valid for classification. 

What I understood from previous designs/conversations is a `Pipeline` is a `Model`+`Data reader`. I think the API could reflect this better.

Maybe is just about naming  and arranging things a bit differently.

dvsrepo:
> > I have added now a `TokenClassifier` example. Backbone model is (almost) equivalent to the TextClassification one, except from the num_wrapping_dim and TimeDistributedEncoder which do not make much sense outside hierarchical classification tasks.
> > Instantiation not working yet due to CRF instantiation failing with num_labels=0 as we have no "outside vocab" when instantiating from_params. Will check this tomorrow morning, next I will prepare example data and test the full pipeline.
> 
> Now it's working. There is a problem with output label key names (i think it's returning the tag label, but i'm not sure)

Let's discuss this tomorrow, this kind of "breaks" the generality/reusability of the models we had for text_classifiers, where we did not have to include the labels in configuration files. Maybe it just have to be this way but would like to check this together.

I have included now a train example with an implementation of the `text_to_instance` method. 

In general, I've found it easy to extend and create a new model.

We should create/use more realistic datasets to actually test if the models are working as expected, I'll do this tomorrow.

dcfidalgo:
Regarding this:
> It is a good chance to review naming and keep it as simple as possible for developers (e.g., we want to keep learn or go to more standard train).

I would vote +1 for the renaming of `learn` -> `train`. And maybe also `features` - > `embedding` in the config file from above.

dcfidalgo:
I will make a list here of things i see going through this PR. This comment is work-in-progress, just to have a place to write down and save my findings. I will try to keep this list up-to-date:

- **src/biome/text/api_new/vocabulary.py**
In general, i think lowercase class names are a bad idea. In the `vocabulary` class case, we could just remove the class level and write all class methods as module methods (-> no need for the `@classmethod` decorators). Then instead of importing the class, one would import the module: `from biome.text.api_new import vocabulary`. 
I also saw an import like this in *api_new/model.py*: `from biome.text.api_new.vocabulary import Vocabulary` that actually imports `allennlp.data.Vocabulary`, which i think is a bit confusing ...
- **src/biome/text/api_new/pipeline.py**
I think the `PipelineHelper` class is a similar case. We could remove the class level and put all its class methods inside a module.

frascuchon:
I will merge this PR and we can go on from master since 0.3.0 is already tagged

PR #124, opened 2020-04-28
Training attributes as configuration class

frascuchon:
This PR define a `TrainerConfiguration`  class used for setup the `pipeline.train` method.

### DIFF
--- a/src/biome/text/api_new/pipeline.py
+++ b/src/biome/text/api_new/pipeline.py
@@ -25,17 +25,15 @@ from fastapi import FastAPI
 
 from biome.text.api_new.modules.encoders import TimeDistributedEncoder
 from biome.text.api_new.modules.heads import TaskHead
-from . import constants, helpers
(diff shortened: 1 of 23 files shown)


src/biome/text/api_new/pipeline.py
```diff
@@ -636,40 +632,82 @@ async def status():
...
+                """
+                required, _ = split_signature_params_by_predicate(
+                    pipeline.head.forward,
+                    lambda p: p.default == inspect.Parameter.empty,
+                )
+                return [p.name for p in required] or [None]
+
+            iterator_config = {
+                "batch_size": config.trainer.batch_size,
+                "max_instances_in_memory": config.trainer.batch_size * 3,
```
dcfidalgo, line 654: Maybe i would also expose the magic number 3 to the config.trainer with the default value 3.
During the DZ project i used an value of `batch_size * 1` for example.

PR #127, opened 2020-04-28
feat/biome bimpm definition

dcfidalgo:
Adds our BiMpm implementation to the new API.

In the test i tried to reflect the new way of writing the configuration file.
I added some comments starting with `# Missing` where i was missing the possibility of some configurations in the new scheme.


src/biome/text/api_new/modules/heads/bimpm_classification.py
```diff
@@ -0,0 +1,487 @@
...
+
+    Parameters
+    ----------
+    model : ``Model``
+        Takes care of the embedding
+    labels : `List[str]`
+        List of labels
+    matcher_word : ``BiMpmMatching``
+        BiMPM matching on the output of word embeddings of record1 and record2.
+    encoder : ``Seq2SeqEncoder``
```
dvsrepo, line 42: I think we should move this encoder to the backbone model, to make this model more consistent and to allow for reusing encoders as we do for other heads.

src/biome/text/api_new/modules/heads/bimpm_classification.py
```diff
@@ -0,0 +1,487 @@
...
+        Encoder layer for record1 and record2
+    matcher_forward : ``BiMPMMatching``
+        BiMPM matching for the forward output of the encoder layer
+    aggregator : ``Seq2VecEncoder``
```
dvsrepo, line 46: We should use specs and limit as much as possible the need of defining input_dim's in config (at least for Seq2Vec and Seq2Seq but I would recommend to do it for every layer if possible

src/biome/text/api_new/modules/heads/bimpm_classification.py
```diff
@@ -0,0 +1,487 @@
...
+        )
+
+        # embedding and encoding of record1
+        embedded_record1 = self.dropout(
+            self.model.forward(
+                record1, mask=mask_record1, num_wrapping_dims=self.num_wrapping_dims
+            )
+        )
+        encoded_record1 = self.dropout(
+            self.td_encoder(embedded_record1, mask=mask_record1)
```
dvsrepo, line 233: When moving this to backbone model we can do something like `self.model.encoder`

tests/text/pipelines/test_multifield_bimpm.py
```diff
@@ -34,59 +33,109 @@ def training_data_yaml(tmpdir):
...
+            #             "type": "gru",
+            #             "input_size": 32,
+            #             "hidden_size": 112,
+            #             "num_layers": 1,
+            #             "bidirectional": True,
+            #         },
+            #         "dropout": 0.2,
+            #     },
+            # },
+            # Missing: lowercase_characters, padding_index
```
dvsrepo, line 74: `lowercase_characters` should be included as a top level config param in `CharFeaturesSpec` as it is sufficiently common, same with others you consider highly relevant. Less frequent configs can still be included in configuration time using `embedder` and `indexer` sections in this block (we will document this)

tests/text/pipelines/test_multifield_bimpm.py
```diff
@@ -96,46 +145,60 @@ def pipeline_yaml(tmpdir):
...
+        #     "batch_size": 3,
+        #     "sorting_keys": [
+        #         [
+        #             "record1",
+        #             "num_fields" if request.param == "multifield" else "num_tokens",
+        #         ]
+        #     ],
+        #     "type": "bucket",
+        # },
+        # Missing: batch_size; max_instances_in_memory; cache_instances;
```
frascuchon, line 168: `batch_size` and `cache_instances` are properly exposes in [this PR](https://github.com/recognai/biome-text/pull/124/files#diff-d550678364915bd06d14ae55044d0fbbR91). Also,  `max_instances_in_memory` is auto-calculated, and should be since could be a failing point in training process.

src/biome/text/api_new/modules/heads/bimpm_classification.py
```diff
@@ -0,0 +1,487 @@
...
+from allennlp.modules.bimpm_matching import BiMpmMatching
+from allennlp.nn import InitializerApplicator
+from allennlp.nn import util
+from biome.text.api_new.model import Model
+from biome.text.api_new.modules.encoders import TimeDistributedEncoder
+from biome.text.api_new.modules.heads.classification.defs import ClassificationHead
+from overrides import overrides
+
+
+class BiMpm(ClassificationHead):
```
frascuchon, line 21: Should be nice implement the `prediction_explain` for enable explains functionality in this kind of model.

You can take a look into [`DocumentClassification`](https://github.com/recognai/biome-text/blob/feat/add_our_biome_bimpm/src/biome/text/api_new/modules/heads/doc_classification.py#L103) and [`TextClassification`](https://github.com/recognai/biome-text/blob/feat/add_our_biome_bimpm/src/biome/text/api_new/modules/heads/text_classification.py#L73) implementations, since should be a merge of both

src/biome/text/api_new/modules/heads/bimpm_classification.py
dcfidalgo: I tried to move it, but it was messy, since the `matcher_word` compares the embeddings, and therefore needs the ouput of the `model.embedder`. I find defining a `PassThroughEncoder` as the `model.encoder` a much cleaner solution for now.

src/biome/text/api_new/modules/heads/bimpm_classification.py
dcfidalgo: Done

src/biome/text/api_new/modules/heads/bimpm_classification.py
dcfidalgo: I used the new `TimeDistributedEncoder` to remove the need for the separate `*.td_encoder`s.

tests/text/pipelines/test_multifield_bimpm.py
dcfidalgo: done

tests/text/pipelines/test_multifield_bimpm.py
```diff
@@ -34,59 +33,109 @@ def training_data_yaml(tmpdir):
...
+        "features": {
+            # "tokens": {
+            #     "indexer": {"type": "single_id", "lowercase_tokens": True},
+            #     "embedder": {
+            #         "type": "embedding",
+            #         "embedding_dim": 32,
+            #         "padding_index": 0,
+            #     },
+            # },
+            # Missing: padding_index
```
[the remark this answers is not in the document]
dcfidalgo: done

src/biome/text/api_new/modules/heads/bimpm_classification.py
dcfidalgo: agreed, would be very nice indeed! I will have a look at it tomorrow.

dcfidalgo:
Added a new commit, in which i addressed most of the inline comments.

tests/text/pipelines/test_multifield_bimpm.py
dcfidalgo: Got it

dcfidalgo:
As internally discussed, i removed the check for the PassThroughEncoder to be able to use pretrained `biome.text.api_new.model.Model`s.

@frascuchon @dvsrepo  Can this go in? Having the tag for DZ, i would also vote for merging [this branch](https://github.com/recognai/biome-text/tree/feature/hierarchical-pipeline-definition) into master, and then in the near future make a cleanup PR, in which we clean up the old stuff and move the new stuff out of the `api_new` folder.

dcfidalgo:
@frascuchon Have a look at commit 9a10b44b7ef492cdbe8d5467aa7612f3e721b8cf . In this way we avoid deepcopy calls for each token, while changes to `self.features` will have no effect. Also, do we need the `self.featurizer` attribute?

frascuchon:
> @frascuchon Have a look at commit [9a10b44](https://github.com/recognai/biome-text/commit/9a10b44b7ef492cdbe8d5467aa7612f3e721b8cf) . In this way we avoid deepcopy calls for each token, while changes to `self.features` will have no effect. 

Grea!. Anyway, i've created a PR that will helps to resolve the problem definitely 
https://github.com/recognai/biome-text/pull/129/files

> Also, do we need the `self.featurizer` attribute?
Mmmm, we use it for some init checks in `language_modelling.py`. I need to think about featurizer, features and embedding to keep a single entrypoint for all operations, but for now, we cannot remove the featurizer.

PR #128, opened 2020-04-30
Feature/calculate frozen and trainable pipeline parameters

frascuchon:


src/biome/text/api_new/pipeline.py
```diff
@@ -497,6 +497,23 @@ def type_name(self) -> str:
         """The pipeline name. Equivalent to task head name"""
         return self.head.__class__.__name__
 
+      
+    def _filter_parameters_by_predicate(
+        self, predicate: Callable[[], bool]
+    ) -> List[str]:
+        filtered_parameters = filter(predicate, self._model.named_parameters())
```
dcfidalgo, line 504: @frascuchon For me `self._model.named_parameters()` returns a tuple, so `p.requires_grad` in the lambda functions fails ... i tested it with pytorch 1.2 and 1.4

src/biome/text/api_new/pipeline.py
```diff
@@ -497,6 +497,23 @@ def type_name(self) -> str:
...
+
+    @property
+    def trainable_parameters(self) -> List[str]:
+        """Returns the name of pipeline trainable parameters"""
+        return self._filter_parameters_by_predicate(lambda p: p.requires_grad)
+
+    @property
+    def frozen_parameters(self) -> List[str]:
+        """Returns the number of frozen parameters of pipeline"""
+        return self._filter_parameters_by_predicate(lambda p: not p.requires_grad)
```
dcfidalgo, line 515: Does `frozen_parameters` return the **number** and `trainable_parameters` the **name** of the parameters? Why not numbers for both?

src/biome/text/api_new/pipeline.py
frascuchon: Yes, it was a refactor error. Fixed now

src/biome/text/api_new/pipeline.py
frascuchon: We just keep info about trainable params

PR #132, opened 2020-05-04
feature: customize allennlp training

frascuchon:
This PR detaches training process from allennlp's `fine_tune` method, customizing some of its functionality:

- Allows restore a checkpointed training process 
- Allows save vocabulary into experiment folder for several uses


src/biome/text/api_new/pipeline.py
```diff
@@ -632,103 +693,90 @@ async def status():
...
+        pipeline._model.launch_experiment(
+            params=Params(allennlp_configuration(pipeline, config)),
             serialization_dir=config.output,
-            extend_vocab=True,  # TODO: Allow parameterize
+            extend_vocab=config.extend_vocab,
             file_friendly_logging=True,
-        )
+        )  # pylint: disable=protected-access,
 
         return pipeline.__class__(
```
dvsrepo, line 781: A bit unrelated, but isn't the config ignored (taken from archive) when constructing the `Pipeline` from pretrained? I had left a comment  in the __init__ method, regarding the signature which is a bit confusing with two Optionals, not sure if it's too relevant though.

frascuchon: Yes, the arguments are exclusives. So we just need to include some argument validations in top of `__init__` for avoid confussions

PR #133, opened 2020-05-05
feat: better vocab extension mechanisms

frascuchon:
This PR helps to :

- load vocab from a given path
- extend loaded vocab in pipeline initialization with selected sources
- keep updated module components that uses vocab in some way for its setup


src/biome/text/api_new/configuration.py
```diff
@@ -213,14 +213,16 @@ def __init__(
 
 
 class VocabularyConfiguration:
-    """Configures a `Vocabulary` before it gets created from data
+    """Configures a ``Vocabulary`` before it gets created from data
 
-    Use this to configure a Vocabulary using specific arguments from `allennlp.data.Vocabulary`
+    Use this to configure a Vocabulary using specific arguments from `allennlp.data.Vocabulary``
```
dvsrepo, line 218: Missing `

src/biome/text/api_new/configuration.py
```diff
@@ -213,14 +213,16 @@ def __init__(
...
 
     See [AllenNLP Vocabulary docs](https://docs.allennlp.org/master/api/data/vocabulary/#vocabulary])
 
     Parameters
     ----------
+    from_path: ``Optional[str]``
+        If provided, try to load model vocab from specified folder path
```
dvsrepo, line 225: If provided, try to load model vocab from specified folder path containing datasource YAML definitions ?

src/biome/text/api_new/modules/heads/language_modelling.py
```diff
@@ -73,6 +73,12 @@ def __init__(self, model: Model, dropout: float = None) -> None:
             embedding_dim=self.model.encoder.get_output_dim(),
         )
 
+    def _on_vocab_update(self):
```
dvsrepo, line 76: Great!

src/biome/text/api_new/model.py
```diff
@@ -50,6 +50,11 @@ def forward(
         embeddings = self._embedder(text, num_wrapping_dims=num_wrapping_dims)
         return self.encoder(embeddings, mask=mask)
 
+    def extend_vocab(self, vocab: Vocabulary, **kwargs):
+        """This method is automatically called when a base model updates the vocabulary"""
+        self.vocab = vocab
+        self._embedder = self.featurizer.build_embedder(self.vocab)
```
dvsrepo, line 56: This will keep the weights of existing embeddings and extend the size of embedding layers, right?

src/biome/text/api_new/configuration.py
frascuchon: No, the folder will contains an already saved vocab

frascuchon: It will use the path for load a vocab using `Vocabulary.from_files` method

src/biome/text/api_new/model.py
frascuchon: This will reload the embedding layer with new vocab, so, every embedding configurations will be the same. 

Maybe we can use `Model.extend_embedder_vocab` in that case

PR #138, opened 2020-05-06
feat: remove old modules

frascuchon:
This PR moves all `api_new` modules to the root package module and removes the unneeded old modules


src/biome/text/pipeline.py
```diff
@@ -215,6 +215,8 @@ class Pipeline:
     __LOGGER = logging.getLogger(__name__)
     __TRAINING_CACHE_DATA = "instances_data"
 
+    logging.getLogger("allennlp").setLevel(logging.WARNING)
+
```
dcfidalgo, line 219: Does this disable the `verbose` option of the `train` command?

frascuchon: You're right. I'll make the changes to preserve the `verbose` feat.

frascuchon: Reverted

PR #144, opened 2020-05-08
refactor: pipeline API & internal deps

### LINKED ISSUE #143: Check pipeline init arguments
Must specify a minimal check for passing exclusive arguments `pretrained_path` and `config` in pipeline init method

frascuchon:
This PR include several changes over pipeline class definition and its inner dependencies.

The main `Pipeline` can only be initialized using the methods:

- `Pipeline.from_yaml`/`Pipeline.from_config`: for load a "blank" pipeline
- `Pipeline.from_pretrained` : for load already pretrained pipeline

There are to new inner `PIpeline` classes  managing the loading cases: `_BlankPipeline` and `PreTrainedPipeline`

Also, the `train`method in a pipeline doesn't not modify the original pipeline, so you can launch several training processes from the same pipeline. 

```python
    train = "train.data.yml"
    validation = "validation.data.yml"
    training_folder = "experiment"

    pl = Pipeline.from_yaml(
        "text_classifier.yaml", vocab_path=os.path.join(training_folder, "vocabulary")
    )

    trainer_configuration = TrainerConfiguration(**yaml_to_dict("trainer.yml"))
    trainer_configuration.data_bucketing = False
    pl.train(
        output=training_folder,
        trainer=trainer_configuration,
        training=train,
        validation=validation,
        extend_vocab=VocabularyConfiguration(sources=[train, validation]),
    )

    trained_pl = Pipeline.from_pretrained(os.path.join(training_folder, "model.tar.gz"))
    
```
Closes #143

### DIFF
--- a/src/biome/text/featurizer.py
+++ b/src/biome/text/featurizer.py
@@ -149,7 +149,7 @@ class InputFeaturizer:
         
         Returns
         -------
-        An `InmutableDict` defining the token indexers of the featurizer
+        An dictionary defining the token indexers of the featurizer
         """
         # fmt: off
         return {
--- a/src/biome/text/featurizer.py
+++ b/src/biome/text/featurizer.py
@@ -152,10 +152,10 @@ class InputFeaturizer:
         An `InmutableDict` defining the token indexers of the featurizer
         """
         # fmt: off
-        return InmutableDict({
+        return {
             feature: TokenIndexer.from_params(Params(config[self.__INDEXER_KEYNAME]))
             for feature, config in self.config.items()
-        })
+        }
         # fmt: on
 
     def build_embedder(self, vocab: Vocabulary) -> Embedder:
--- a/documentation/docs/documentation/02.tutorials/01.Training_a_text_classifier.md
+++ b/documentation/docs/documentation/02.tutorials/01.Training_a_text_classifier.md
@@ -316,7 +316,7 @@ head:
 
 
 ```python
-pl = Pipeline.from_file("configs/text_classifier.yml")
+pl = Pipeline.from_yaml("configs/text_classifier.yml")
 ```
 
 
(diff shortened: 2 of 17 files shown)


dvsrepo:
I agree on simplifying the Init and instantiation methods. Init only with
only pipeline configuration and then from_pretrained and from_file (or from
YAML).

I prefer from pretrained as we have it now, because it's not only a binary
its more things, the vocab etc.

frascuchon:
I think we are just mixing concepts. For me, a PipelineConfiguration keep info about how the pipeline is defined. So, including info about where is archived, how is loaded, or how to build a vocab, are parametres out of definition. Even more, moving forward the pretrained_path into configuration, doesnt resolve the initial ambiguity of initialization.

There are some other parametres nota cover here, like use GPU device that we wouldnt like to include as pipeline configuration ( i Hope)

dvsrepo:
On my side, the only change I was talking about is to remove the parameter
pretrained path in the init, because it creates a confusing signature. If
you want to instantiate a pipeline from archive you'd use the
from_pretrained. I think that indeed removes the ambiguity.

dcfidalgo:
i made a commit with a proposal, but as i said, i'm also more than fine with the original PR!
One thing that i find a bit confusing concerning names is this: `pipeline._model.head.model`
`_model` and `model` are different things but they share the same name ... i guess the second `model` is rather only the "encoder" + "embedder" in contrast to the `_model` which holds all of the weights, is this true?

src/biome/text/featurizer.py
```diff
@@ -152,10 +152,10 @@ def build_features(self) -> Dict[str, TokenIndexer]:
         An `InmutableDict` defining the token indexers of the featurizer
```
dvsrepo, line 152: Update docs

frascuchon: Done!

PR #148, opened 2020-05-12
refactor: rename model to backbone encoder

### LINKED ISSUE #147: Rename Model class to BackboneEncoder

frascuchon:
Close #147

### DIFF
--- a/src/biome/text/backbone.py
+++ b/src/biome/text/backbone.py
@@ -12,8 +12,8 @@ from .tokenizer import Tokenizer
 from .vocabulary import Vocabulary
 
 
-class Model(torch.nn.Module):
-    """Model definition. All models used in pipelines must configure this model class"""
+class BackboneEncoder(torch.nn.Module):
+    """Backbone Encoder definition. All models used in pipelines must configure this model class"""
 
     def __init__(
         self,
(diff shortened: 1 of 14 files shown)


src/biome/text/modules/heads/defs.py
```diff
@@ -58,9 +58,9 @@ def as_string(self) -> str:
 class TaskHead(torch.nn.Module, Registrable):
     """Base task head class"""
 
-    def __init__(self, model: Model):
+    def __init__(self, backbone: BackboneEncoder):
         super(TaskHead, self).__init__()
-        self.model = model
+        self.backbone = backbone
```
dcfidalgo, line 63: Here i would also be more specific: `self.backbone` -> `self.backbone_encoder`
Here i think it is important, since this avoids a "what the fuck is a backbone?" moment in all the docstrings of the heads (TextClassification, LanguageModeling, ...)

frascuchon: The backbone encoder should be described at some early point in documentation. So, at this point, users should be sufficiently familiar with the concepts of the library. Isn't it?

src/biome/text/backbone.py
```diff
@@ -12,8 +12,8 @@
 from .vocabulary import Vocabulary
 
 
-class Model(torch.nn.Module):
-    """Model definition. All models used in pipelines must configure this model class"""
+class BackboneEncoder(torch.nn.Module):
```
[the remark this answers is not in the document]
frascuchon: I keep the PEP 8 style guides for python code related to module names:

> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

PR #161, opened 2020-05-13
fix mask for the encoding after the pooling over the TextFields

dcfidalgo:
This PR concerns the mask in the `DocumentationClassification` after reducing the dimension by means of the `tokens_pooler`.

I stumbled across this one while designing the new BiMpm for records. The method we used so far (`get_text_field_mask`) only checks if the value of the tensor is `!= 0`, which is designed for TextField tensors with token ids. For tensors coming out of an embedder/encoder this is not reliable!

The new approach just checks if all tokens in a TextField are padding tokens and masks this TextField accordingly.


dvsrepo:
Good catch and good solution, I always felt a bit uncomfortable using the `get_text_field_mask` in that second step.

PR #184, opened 2020-05-18
refactor(config): using words and chars config classes instead of dicts

### LINKED ISSUE #179: InputFeaturizer initialization
Create `InputFeaturizer` objects passing directly words and chars specs, instead of configuration dicts.

```python
InputFeaturizer(
  words=WordsFeaturesSpecs(embedding_dim=50,lowercase_tokens=True),
  chars=CharsFeaturesSpec(embedding_dim=10, encoder={ "type": ....}...)
)
```

Also, align spec naming  and make them public classes

`_CharacterFeaturesSpec` -> `CharsFeaturesSpec`
`_WordFeaturesSpecs` -> `WordsFeaturesSpecs`

frascuchon:
Closes #179

### DIFF
--- a/src/biome/text/_model.py
+++ b/src/biome/text/_model.py
@@ -366,7 +366,7 @@ class PipelineModelTrainer:
 
         os.makedirs(self._serialization_dir, exist_ok=True)
 
-        serialization_params = deepcopy(self._params).as_dict(quiet=True)
+        serialization_params = sanitize(deepcopy(self._params).as_dict(quiet=True))
         with open(
             os.path.join(self._serialization_dir, CONFIG_NAME), "w"
         ) as param_file:
--- a/src/biome/text/_model.py
+++ b/src/biome/text/_model.py
@@ -95,13 +95,15 @@ class PipelineModel(allennlp.models.Model, allennlp.data.DatasetReader):
         if not isinstance(config, PipelineConfiguration):
             config = PipelineConfiguration.from_params(config)
 
+        vocab = vocab or vocabulary.empty_vocab(features=config.features.keys)
+
         return cls(
             name=config.name,
             head=config.head.compile(
                 backbone=BackboneEncoder(
-                    vocab=vocab or vocabulary.empty_vocab(features=config.features.keys),
+                    vocab,
                     tokenizer=config.tokenizer.compile(),
-                    featurizer=config.features.compile(),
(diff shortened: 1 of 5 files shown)


frascuchon:
> * Does this cover the comment from @dcfidalgo regarding alignment of yaml names and class names?

`WordFeaturesConfiguration` doesn't match with @dcfidalgo proposal (`words` key), `word` feature renaming or` WordsFeatureConfiguration`  does.  Is `word` feature naming  enough for a word level features definition ?  (Same for char level features)

> Finally, I think I would prefer `WordFeaturesConfiguration` and `CharFeaturesConfiguration`, as I think they are better aligned with FeaturesConfiguration and other configurations. I do not know if there's a strong reason to call them Specs, but I think it adds some (unnecessary) cognitive load (even if we do some sort of `compile` inside the code I think is an implementation detail and not needed for the user-facing API).

+1. My background make me sometimes create to technical names or too general software engineering oriented.

dvsrepo:
> > * Does this cover the comment from @dcfidalgo regarding alignment of yaml names and class names?
> 
> `WordFeaturesConfiguration` doesn't match with @dcfidalgo proposal (`words` key), `word` feature renaming or` WordsFeatureConfiguration` does. Is `word` feature naming enough for a word level features definition ? (Same for char level features)
This is a good question, I think `features: ['word': ...., 'char': ...]` works well also in YAML files but I don't have a strong opinion.

> 
> > Finally, I think I would prefer `WordFeaturesConfiguration` and `CharFeaturesConfiguration`, as I think they are better aligned with FeaturesConfiguration and other configurations. I do not know if there's a strong reason to call them Specs, but I think it adds some (unnecessary) cognitive load (even if we do some sort of `compile` inside the code I think is an implementation detail and not needed for the user-facing API).
> 
> +1. My background make me sometimes create to technical names or too general software engineering oriented.
> 
> If we are ok, we should tackle this in another PR include all `*Spec` renaming (task, component, ...)

+1 to make another PR to include all `Spec` renaming, I was ok with Spec but I realise this seeing all the configurations together and developing some of the examples.

frascuchon:
@dvsrepo I've changed `words` to `word` references (same for `chars`) and rename specs to `WordFeatures` and `CharFeatures` 

We can discuss about `*Configuration`  naming in another PR.

The PR is ok for review

dvsrepo:
Perfect!

PR #189, opened 2020-05-19
wip: AllenNLP 1.0 integration

dvsrepo:
Ongoing issues:

- Custom batch_samplers are not allowed for IterableDatasets in Pytorch DataLoaders so we cannot use BucketingSampler.

- Other remaining issue is text_cleaning registration, which also fails, but I remove this config from example1 for now as I think that should be a minor issue.


examples/1.text_classifier/text_classifier.yaml
```diff
@@ -1,9 +1,9 @@
 name: text-classifier
 
-tokenizer:
-    text_cleaning:
-        rules:
-            - strip_spaces
+#tokenizer:
+#    text_cleaning:
```
dvsrepo, line 4: @frascuchon @dcfidalgo this is not working with 1.0, seems like an issue with nested registrable objects. Besides this is the only place where we use annotations to register components.

src/biome/text/_model.py
```diff
@@ -377,8 +376,18 @@ def _setup(self):
...
         vocab.save_to_files(os.path.join(self._serialization_dir, "vocabulary"))
 
-        self._iterator = DataIterator.from_params(self._params.pop("iterator"))
-        self._iterator.index_with(vocab)
+        #self._iterator = DataIterator.from_params(self._params.pop("iterator"))
+        #self._iterator.index_with(vocab)
+        for dataset in self._all_datasets.values():
+            dataset.index_with(vocab)
+            
+        self._dataloader = DataLoader(dataset=self._all_datasets["train"])
```
dvsrepo, line 384: @frascuchon here I am ignoring data_bucketing (iterator)

src/biome/text/_model.py
```diff
@@ -377,8 +378,29 @@ def _setup(self):
...
+        #self._iterator = DataIterator.from_params(self._params.pop("iterator"))
+        #self._iterator.index_with(vocab)
+        iterator_config = self._params.pop("iterator")
+        
+        for dataset in self._all_datasets.values():
+            dataset.index_with(vocab)
+        
+        self._dataloader = DataLoader(
+            dataset=self._all_datasets["train"],
+            batch_size=iterator_config["batch_size"]
```
dvsrepo, line 390: @frascuchon @dcfidalgo 
I included now the batch_size. 

PyTorch DataLoaders do not allow custom batch_samplers (for using previous iterator params for bucketing) for IterableDatasets, this is a known issue  https://github.com/allenai/allennlp/issues/3966 but frankly I think we can live without bucketing for now for the type of problems we are tackling. We can maybe discuss this later today.

src/biome/text/configuration.py
```diff
@@ -160,7 +160,6 @@ class TrainerConfiguration:
     optimizer
     validation_metric
     patience
-    shuffle
```
dvsrepo, line 163: this param has been removed from Trainer configuration for 1.0. Same thing as with batch_sampler DataLoaders https://github.com/allenai/allennlp/issues/3966

dvsrepo:
I think we can close this one

PR #195, opened 2020-05-19
New RecordPairClassification Head

dcfidalgo:
This PR introduces the new BiMpm (`RecordBiMpm`) adapted to "record-like" data, that is a record consisting of several data fields.

I think overall it is simpler to understand:
- less parameters
- clear distinction between the two encoders (`field_encoder` = seq2vec <-> `record_encoder` = seq2seq) 
- only the output of the `record_encoder` is passed on to the matching layer

It performs really solid, i tested it with the Uniserv Datapitch data: https://gitlab.com/recognai-team/biome/biome-classifiers/-/blob/master/experiments/record_bimpm

This version also allows me to include a meaningful calculation of attributions for each data field by means of the integrated gradients method. An example can be seen here:
https://gitlab.com/recognai-team/biome/biome-classifiers/-/blob/master/experiments/record_bimpm/record_bimpm_example.ipynb
(somehow it does not show the colors in the visualization of the attributes, so pretty useless. You have to download the notebook to see it).

There are several things that i would like to discuss briefly in a call:
- How to deal with *GoldenRecords*, that is 2-level data structures
- exact format of the output of `explain` (this is related to the whole explain UI stuff)
- target for attribute calculation in case of binary classifiers
- csv vs json when providing dicts as input data

@dvsrepo @frascuchon comments are more than welcome!

TODO:
- rename to `RecordPairClassification` **Done**
- define final `explain_prediction` output


dcfidalgo:
TODO:
- rename to `RecordPairMatcher` or `RecordPairClassification`

src/biome/text/_model.py
```diff
@@ -288,7 +288,6 @@ def _read(self, file_path: str) -> Iterable[Instance]:
         """An generator that yields `Instance`s that are fed to the model
```
dvsrepo, line 288: A generator

src/biome/text/modules/heads/__init__.py
```diff
@@ -1,11 +1,11 @@
 # fmt: off
-from .bimpm_classification import BiMpm
 from .defs import TaskHead, TaskHeadSpec, TaskName, TaskOutput
 from .doc_classification import DocumentClassification
 from .language_modelling import LanguageModelling, LanguageModellingSpec
 from .record_classification import RecordClassification, RecordClassificationSpec
 from .text_classification import TextClassification, TextClassificationSpec
 from .token_classification import TokenClassification, TokenClassificationSpec
+from .record_bimpm_classification import RecordBiMpm
```
dvsrepo, line 8: As you mentioned `RecordPairClassification` is a good name for this.

src/biome/text/modules/heads/record_bimpm_classification.py
```diff
@@ -0,0 +1,550 @@
...
+    FeedForwardSpec,
+)
+from captum.attr import IntegratedGradients
+
+from biome.text.featurizer import WordFeatures, CharFeatures
+
+
+class RecordBiMpm(ClassificationHead):
+    """
+    This `Model` is loosely based on the AllenNLP implementation of the BiMPM model described in
```
dvsrepo, line 31: I would start the docstring with a single sentence like Classifiers pairs of records using matching functions...or whatever.

Maybe change this `Model` for this `TaskHead`.

I would add a user-oriented description such as:

Use this head fo problems where you need to classify pairs of records..bla... bla bla. The inputs for models using this head are: (two records? etc.)

src/biome/text/modules/heads/__init__.py
dcfidalgo: Done

src/biome/text/_model.py
dcfidalgo: Done

src/biome/text/modules/heads/record_bimpm_classification.py
dcfidalgo: Done. @dvsrepo If you have further comments for the new doc string please let me know!

dcfidalgo:
Ok, this is ready to merge.
The explain now uses a variable target (depending on the prediction) and i removed the deltas from the output.

PR #204, opened 2020-05-20
Move probabilities calculation to decode

dcfidalgo:
This PR moves the calculation of the probabilities and other UI stuff to the `decode` method, that is only called for predictions.

@frascuchon I left one TODO about the duplicated information in `output.label` and `output.max_class`. I guess removing one or the other has implications for the UI, but we should only use one of the keys.

### DIFF
--- a/src/biome/text/modules/heads/classification/defs.py
+++ b/src/biome/text/modules/heads/classification/defs.py
@@ -70,37 +70,33 @@ class ClassificationHead(TaskHead):
     def task_name(self) -> TaskName:
         return TaskName.text_classification
 
-    def process_output(self, output: TaskOutput) -> TaskOutput:
-        def labels_with_probabilities(probabilities: torch.Tensor) -> Dict[str, float]:
-            """
-            Calculates the descendant sorted label + probs dictionary
-            using all output classes (not only predicted)
-            """
-            all_classes_probs = torch.zeros(
-                self.num_labels,
-                device=probabilities.get_device()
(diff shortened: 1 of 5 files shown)


src/biome/text/modules/heads/classification/defs.py
```diff
@@ -112,10 +108,32 @@ def labels_with_probabilities(probabilities: torch.Tensor) -> Dict[str, float]:
         output.max_class = max_classes
         output.max_class_prob = max_classes_prob
 
+        # TODO: This is just a duplicate of output.max_class. Revise if it is still necessary!
         output.label = max_classes
         output.prob = max_classes_prob
 
-        return output
+        return output.as_dict()
```
frascuchon, line 115: Just a minimal suggestion. Could we do the taskOutput -> dict transformation outside this method (we could do it in `PipelineModel.decode`.

We can loose control over the head output with dictionaries

dcfidalgo: :+1: i like, Done

dcfidalgo:
@frascuchon What about the duplicated information in `output.label` and `output.max_clas`s? Leave it for another PR?

frascuchon:
> @frascuchon What about the duplicated information in `output.label` and `output.max_class`? Leave it for another PR?

Yes, we need to abord this change in UI too. We could just set at deprecated the `max_class` and `max_prob` properties

PR #209, opened 2020-05-21
feat(train): check empty vocabs

frascuchon:


src/biome/text/pipeline.py
```diff
@@ -192,7 +194,13 @@ def train(
                 vocab = self.extend_vocabulary(model.vocab, vocab_config=extend_vocab)
             if vocab:
                 model.set_vocab(vocab)
-            # TODO: empty vocab check
+
+            if vocabulary.is_empty(model.vocab, self.config.features.keys):
+                raise EmptyVocabError(
+                    "Found an empty vocabulary. "
+                    "Please load a vocabulary or define the vocab extension properly"
```
dvsrepo, line 201: would it be possible to add a bit more guidance:

Please load a vocabulary using `(what param/method use for this)` or  define the vocab extension with  the `extend_vocab` parameter using a `VocabularyConfiguration` object.

PR #216, opened 2020-05-22
integration test for the text classification

dcfidalgo:
This PR adds an integration test using the `TextClassification` head.

On my machine it takes <1 min and the numbers are reproducible.
It covers only a small part of the functionality, but with this test we would have caught the embedding bug for example.
The idea is that with time we extend the test to cover more functionalities, and maybe it can serve as blue print for other integration tests.

### DIFF
--- a/tests/integration/test_text_classification.py
+++ b/tests/integration/test_text_classification.py
@@ -9,6 +9,8 @@ import torch
 import yaml
 from biome.text import Pipeline, TrainerConfiguration, VocabularyConfiguration
 from biome.text.data import DataSource
+from biome.text.configuration import WordFeatures
+from biome.text.configuration import CharFeatures
 
 
 @pytest.fixture
@@ -94,6 +96,7 @@ def trainer_dict() -> dict:
 def test_text_classification(
     tmp_path, pipeline_dict, trainer_dict, train_valid_data_source
 ):
+    """Apart from a well specified training, this also tests the vocab creation!"""
     random.seed(42)
     np.random.seed(422)
     torch.manual_seed(4222)
@@ -107,6 +110,9 @@ def test_text_classification(
     )
 
     pl.create_vocabulary(vocab)
+
+    assert pl._model.vocab.get_vocab_size(WordFeatures.namespace) == 52
+    assert pl._model.vocab.get_vocab_size(CharFeatures.namespace) == 83
 
     output = tmp_path / "output"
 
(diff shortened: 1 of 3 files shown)


dcfidalgo:
Just realized that it this is useless for the CI, since i get the data from our S3 ... maybe we can discuss briefly where to put this data, since it is the same as for our tutorials.

tests/integration/test_text_classification.py
```diff
@@ -0,0 +1,137 @@
...
+    train = create_csv_and_return_yaml(df_train, "train")
+    valid = create_csv_and_return_yaml(df_valid, "valid")
+
+    return str(train), str(valid)
+
+
+@pytest.fixture
+def pipeline_dict():
+    """Pipeline config dict. You need to update the labels!"""
+    pipeline_dict = {
```
frascuchon, line 38: It would be nice to programatically create the pipeline configuration `PipelineConfiguration` and the save as a yaml file. 

We can detect errors related with attributes renaming/deleting or refactoring

tests/integration/test_text_classification.py
```diff
@@ -0,0 +1,137 @@
...
+            },
+        },
+    }
+
+    return pipeline_dict
+
+
+@pytest.fixture
+def trainer_dict():
+    return {
```
frascuchon, line 99: Same comment here

frascuchon:
> Just realized that it this is useless for the CI, since i get the data from our S3 ... maybe we can discuss briefly where to put this data, since it is the same as for our tutorials.

I prefer keep a local copy of data used in tests. Some metrics could change if we update the original data.

dvsrepo:
We can keep a very small local sample of the data and assert that the model
overfits.

tests/integration/test_text_classification.py
dcfidalgo: > It would be nice to programatically create the pipeline configuration PipelineConfiguration and the save as a yaml file.

This was my first approach, but i did not manage. I will try again today.

tests/integration/test_text_classification.py
dcfidalgo: oks

dcfidalgo:
> We can keep a very small local sample of the data and assert that the model
overfits.

The idea was to be able to detect even subtle changes that maybe go undetected when completely overfitting a small data set. Maybe the question is to find a good balance between data set size and being able to catch subtle changes. 
@frascuchon What is the maximum size of the data set i should include as local copy in the test?

frascuchon:
> @frascuchon What is the maximum size of the data set i should include as local copy in the test?

Your set dataset could be enough since take no quite training time. If I remember was set in code to 2000 records, isn't it ?

dcfidalgo:
Ok, i think this can go in.
@frascuchon i added a few asserts regarding the vocab, not sure if you had some more thorough checks in mind. We could add them here or in a follow-up PR.
> It would be nice to programatically create the pipeline configuration PipelineConfiguration and the save as a yaml file.
@frascuchon If you are ok with it, i will add this test in a specific `def test_pipeline_configuration` in another PR, in order not to delay further this one ...

PR #218, opened 2020-05-24
refactor: allow create and load vocabulary for pipelines from scratch

frascuchon:
This PR defines a better vocabulary creation flow. Includes two method to `Pipeline` class for this purpose.

For create or load a vocabulary for a "from-scratch" pipeline you can do as follow:

```python
pipeline = Pipeline.from_yaml("your_config.yaml")
pipeline.create_vocabulary(vocab_config=VocabularyConfiguration(....))
# or, if you want to use an already created vocab
# pipeline.load_vocabulary("path/to/your/vocab/folder")

pipeline.train(....)
```

The vocab creation/load is not allowed for pretrained pipelines, raising an error.

Also, this flow restrict the vocab extension config (in training) only to pretrained pipelines, raising an error if you tries to extend a vocab for an "from-scratch" pipeline.

### DIFF
--- a/examples/1.text_classifier/text_classifier.py
+++ b/examples/1.text_classifier/text_classifier.py
@@ -12,6 +12,7 @@ if __name__ == "__main__":
     training_folder = "experiment"
 
     pl = Pipeline.from_yaml("text_classifier.yaml")
+
     print(pl.predict(text="Header main. This is a test body!!!"))
 
     pl.create_vocabulary(
--- a/examples/1.text_classifier/text_classifier.py
+++ b/examples/1.text_classifier/text_classifier.py
@@ -1,6 +1,9 @@
 import os
 
-from biome.text import Pipeline, VocabularyConfiguration, TrainerConfiguration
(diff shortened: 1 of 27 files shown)


dvsrepo:
Good morning! looks good!

I have added a failing config.yml to the text_classifier example which I found this weekend during testing related to pretrained embeddings. This seem a bit tricky as there is no "real" vocab at the stage where pretrained embeddings are loaded and therefore the error message. Maybe @frascuchon has some more insights on this. We can discuss it today.

frascuchon:
> Good morning! looks good!
> 
> I have added a failing config.yml to the text_classifier example which I found this weekend during testing related to pretrained embeddings. This seem a bit tricky as there is no "real" vocab at the stage where pretrained embeddings are loaded and therefore the error message. Maybe @frascuchon has some more insights on this. We can discuss it today.

I've tested the configuration in master branch and I can reproduce the error. It's not a problem related to this PR. We will tackle this problem in a separated branch.

@dvsrepo  Could you rollback the changes?

PR #219, opened 2020-05-24
Upgrade to allennlp 1.0

frascuchon:
This PR applies analysis described [here](https://github.com/recognai/biome-text/pull/189) for a fully functional allennlp 1.0 integration.

Current version of allennlp is a release candidate, so, changes could be required.

We need integrate [this PR](https://github.com/recognai/biome-text/pull/218) before apply this PR

### DIFF
--- a/src/biome/text/configuration.py
+++ b/src/biome/text/configuration.py
@@ -81,8 +81,10 @@ class FeaturesConfiguration(FromParams):
         return TextFieldEmbedder.from_params(
             Params(
                 {
-                    feature: config["embedder"]
-                    for feature, config in configuration.items()
+                    "token_embedders": {
+                        feature: config["embedder"]
+                        for feature, config in configuration.items()
+                    }
                 }
             ),
             vocab=vocab,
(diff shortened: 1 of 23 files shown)


frascuchon:
This PR is now ready to merge !!!

src/biome/text/configuration.py
```diff
@@ -244,7 +246,6 @@ def __init__(
         optimizer: Dict[str, Any],
         validation_metric: str = "-loss",
         patience: Optional[int] = None,
-        shuffle: bool = True,
```
dvsrepo, line 247: we should remove shuffle from docstring.

src/biome/text/configuration.py
```diff
@@ -260,7 +261,6 @@ def __init__(
         self.optimizer = optimizer
         self.validation_metric = validation_metric
         self.patience = patience
-        self.shuffle = shuffle
         self.num_epochs = num_epochs
         self.cuda_device = cuda_device
         self.grad_norm = grad_norm
```
dvsrepo, line 266: As I understand, iterator params, except batch_size will have no effect for now. I am ok to leave them for now, and discuss what to do about lazy vs. normal dataset readers, we just need to remember to clean up.

frascuchon: For now, no bucketing is enabled during training phase. 

We can discuss this behaviour in another PR since there is no easy solution

PR #226, opened 2020-05-26
feat(record-pair): allow compare (explain) record with missing keys

frascuchon:
This PR includes a minimal change to `RecordPairClassification` record featurize for even both record before generate features.

This change make available explain over records with missing keys

### DIFF
--- a/src/biome/text/helpers.py
+++ b/src/biome/text/helpers.py
@@ -2,12 +2,15 @@ import inspect
 import os
 import re
 from inspect import Parameter
-from typing import Any, Callable, Dict, List, Tuple, Type
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type
 
+import torch
 import yaml
+from allennlp.data import TextFieldTensors
 from elasticsearch import Elasticsearch
 
 from . import environment
+from .features import CharFeatures, WordFeatures
 
 _INVALID_TAG_CHARACTERS = re.compile(r"[^-/\w\.]")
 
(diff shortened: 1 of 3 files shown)


dcfidalgo:
Thanks for the enhancement @frascuchon !
However, i want to mention 2 things that i think we should keep in mind:
1. This change breaks the ability to classify "non-symmetric" records, that is classifying the relation between two record types that do not share the same keys. For now we only use it for "duplicate"/"not_duplicate" relations, but in general it could classify any kind of relation between two records.
2. To get a better performance, i think you should still equalize the records in a preprocess using an "empty token". This allows you to preserve the order of the data fields/keys, which the model takes into account. I think the model could also manage without a specific order, but it will definitely make convergence harder.

frascuchon:
> 1. This change breaks the ability to classify "non-symmetric" records, that s classifying the relation between two record types that do not share the same keys. For now we only use it for "duplicate"/"not_duplicate" relations, but in general it could classify any kind of relation between two records.

In those cases where records, the `explain_prediction` will not work in any case. What it will be the `record1` becomes to `record2` interpretation for better understand the model? Is there a sense for other classification than record comparison ? 

> 2. To get a better performance, i think you should still equalize the records in a preprocess using an "empty token". This allows you to preserve the order of the data fields/keys, which the model takes into account. I think the model could also manage without a specific order, but it will definitely make convergence harder.

We can discuss later this point. For me the "Empy token" it should be configured in a general way in the tokenizer part.

frascuchon:
@dcfidalgo could we include this special behaviour only for explain predictions?

dcfidalgo:
> @dcfidalgo could we include this special behaviour only for explain predictions?
@frascuchon we then should make sure that the "empty token" is the same for the training data set and the prediction input ... maybe not so easy to implement.

frascuchon:
closing pr

PR #251, opened 2020-06-01
update dependencies

frascuchon:
Just a general dependencies update. Maybe some upgrade could bring errors to dependant packages. We must check it.


dcfidalgo:
i want to check the new dependencies and also try out python 3.8
merging this now

PR #258, opened 2020-06-04
first tutorial with test

dcfidalgo:
This PR introduces a first (incomplete) tutorial, and all the infrastructure for the tutorials:
- automatic build for the documentation
- inclusion to the test suite

The tutorial goes until the `Pipeline.explore()` command, which for me is not working at the moment.

Apart from the tutorial stuff, there are also some minor changes included in this PR (see individual commits):
 - pass on generator instead of list when creating vocab
- add warning to `trainable_parameters` when vocab is empty
- add a `lowercase_characters` argument in the `CharFeatures` class
- the two folders *.instances_data* and *.datasources* in the training output are now hidden folders, they are not really interesting for the user
- add the default mapping for data sources in the explore command, if no mapping is present in the data source


dcfidalgo:
> Only comment would be to remove the training output from the saved state of the notebook, as it is too long

Good point, done!

dcfidalgo:
Just fixed the links in the notebook, this should be good to go.

PR #262, opened 2020-06-04
remove output folder if not `restore`

dcfidalgo:
@frascuchon Is the `__prepare_experiment_folder` method still necessary?
I found that if i wanted to train the model a little bit more with `restore=True`, all the metrics from the first training run were gone. Could we just delete the output folder if restore=False and keep it if restore=True?

### DIFF
--- a/src/biome/text/pipeline.py
+++ b/src/biome/text/pipeline.py
@@ -3,6 +3,7 @@ import glob
 import inspect
 import logging
 import os
+import shutil
 import tempfile
 import uuid
 from inspect import Parameter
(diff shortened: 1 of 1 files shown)


frascuchon:
> @frascuchon Is the `__prepare_experiment_folder` method still necessary?
> I found that if i wanted to train the model a little bit more with `restore=True`, all the metrics from the first training run were gone. Could we just delete the output folder if restore=False and keep it if restore=True?

The restore behaviour have changed a bit from first implementation. Yes, we can just delete the experiment folder

Thanks!

PR #264, opened 2020-06-05
revise doc strings

dcfidalgo:
This PR is a first partial revision of our doc strings. There is still a lot to revise, but i will submit PRs step by step, not to end up with a giant PR.

Apart from the revision, i removed the `skip_empty_tokens` argument for now, since we are not using it at the moment.

One small suggestion, maybe the `Tokenizer` could get as input a `TokenizerConfiguration` object. Otherwise we are repeating the arguments and doc strings for both classes. @frascuchon what do you think? If you are ok with it, i will work on the refactoring.


src/biome/text/configuration.py
```diff
@@ -135,11 +151,15 @@ def _make_allennlp_config(self) -> Dict[str, Any]:
...
-    """Creates a `Tokenizer` configuration
+    """Configures the `Tokenizer`
+
+    For a description of the parameters see `biome.text.tokenizer.Tokenizer`
 
     Parameters
     ----------
     lang
+        The [model used](https://spacy.io/api/tokenizer) for tokenization is language dependent.
+        For optimal performance, specify the language of your input data (default: "eng").
```
dvsrepo, line 162: The [spaCy model used]
- default "en"

src/biome/text/pipeline.py
```diff
@@ -253,9 +253,9 @@ def train(
             allennlp_logger.setLevel(logging.ERROR)
 
     def predict(self, *args, **kwargs) -> Dict[str, numpy.ndarray]:
-        """Predicts over some input data with current state of the model
+        """Returns a prediction given some input data based on the current state of the model
 
-        The accepted input is dynamically calculated from head.input
+        The accepted input is dynamically calculated from `self.head.input`
```
dvsrepo, line 258: can we add a tip/example to check the signature of predict using self.head.input?

src/biome/text/pipeline.py
```diff
@@ -354,60 +360,62 @@ def explore(
         return explore_df
 
     def serve(self, port: int = 9998):
-        """Launches a REST prediction service with current model in a specified port (default is `9998)
+        """Launches a REST prediction service with the current model on a specified port (default is `9998)
```
dvsrepo, line 363: missing `

src/biome/text/pipeline.py
```diff
@@ -418,20 +426,19 @@ def type_name(self) -> str:
     @property
     def trainable_parameters(self) -> int:
         """
-        Return the number of trainable parameters.
-
-        This number could be change before an after a training process, since trainer could fix some of them.
+        Returns the number of trainable parameters.
 
+        Keep in mind that this number also depends on the configuration of your trainer.
```
dvsrepo, line 431: At training time, this number can change when freezing/unfreezing certain parameter groups.

(internal note: although I think this is not possible at the moment)

src/biome/text/configuration.py
dcfidalgo: Done

src/biome/text/pipeline.py
dcfidalgo: i saw that the signature is calculated from the `self.inputs` attribute (property), so i simply added a print statement as small example.

src/biome/text/pipeline.py
dcfidalgo: moved the default to the Parameters section

src/biome/text/pipeline.py
dcfidalgo: Done

dvsrepo:
LGTM

PR #265, opened 2020-06-08
fix: text cleaning and max sequence length before other tokenization processes

dvsrepo:
Text cleaning and max sequence length must be applied before any other process (sentence segmentation, etc.) 

Otherwise it causes issues with preproc functions (segmented html which is not cleaned by html_text) and leads to difficult to understand interactions (e.g., max seq len is only applied at the sentence level).

### DIFF
--- a/src/biome/text/text_cleaning.py
+++ b/src/biome/text/text_cleaning.py
@@ -101,4 +101,5 @@ def fix_html(text: str) -> str:
 @TextCleaningRule
 def html_to_text(text: str) -> str:
     """Extracts text from an HTML document"""
+    print(BeautifulSoup(text, "lxml").get_text())
     return BeautifulSoup(text, "lxml").get_text()


frascuchon:
> Text cleaning and max sequence length must be applied before any other process (sentence segmentation, etc.)
> 
> Otherwise it causes issues with preproc functions (segmented html which is not cleaned by html_text) and leads to difficult to understand interactions (e.g., max seq len is only applied at the sentence level).



> Text cleaning and max sequence length must be applied before any other process (sentence segmentation, etc.)
> 
> Otherwise it causes issues with preproc functions (segmented html which is not cleaned by html_text) and leads to difficult to understand interactions (e.g., max seq len is only applied at the sentence level).

@dvsrepo  Setting a sequence limit could affect t sentence segmentation, isn't it ?

dvsrepo:
> > Text cleaning and max sequence length must be applied before any other process (sentence segmentation, etc.)
> > Otherwise it causes issues with preproc functions (segmented html which is not cleaned by html_text) and leads to difficult to understand interactions (e.g., max seq len is only applied at the sentence level).
> 
> > Text cleaning and max sequence length must be applied before any other process (sentence segmentation, etc.)
> > Otherwise it causes issues with preproc functions (segmented html which is not cleaned by html_text) and leads to difficult to understand interactions (e.g., max seq len is only applied at the sentence level).
> 
> @dvsrepo Setting a sequence limit before sentence segmentation could bring unexpected results

Only for the last sentence of the full text if it gets cut, but we really want to control the full text length before anything else as this will cause potentially very large texts to go downstream into tokenization, sentence splitting etc. Max sequence length should be used to avoid extremely large text chunks. Max nr of sentences can provide a more fine grained measure for size of text.

frascuchon:
> > > Text cleaning and max sequence length must be applied before any other process (sentence segmentation, etc.)
> > > Otherwise it causes issues with preproc functions (segmented html which is not cleaned by html_text) and leads to difficult to understand interactions (e.g., max seq len is only applied at the sentence level).
> > 
> > 
> > > Text cleaning and max sequence length must be applied before any other process (sentence segmentation, etc.)
> > > Otherwise it causes issues with preproc functions (segmented html which is not cleaned by html_text) and leads to difficult to understand interactions (e.g., max seq len is only applied at the sentence level).
> > 
> > 
> > @dvsrepo Setting a sequence limit before sentence segmentation could bring unexpected results
> 
> Only for the last sentence of the full text if it gets cut, but we really want to control the full text length before anything else as this will cause potentially very large texts to go downstream into tokenization, sentence splitting etc. Max sequence length should be used to avoid extremely large text chunks. Max nr of sentences can provide a more fine grained measure for size of text.

Can't we control the text size by combining max sequence length and max number of sentences?

If not, it's hard to see how we apply that maximum sequence length to record-type data.

dvsrepo:
> > > > Text cleaning and max sequence length must be applied before any other process (sentence segmentation, etc.)
> > > > Otherwise it causes issues with preproc functions (segmented html which is not cleaned by html_text) and leads to difficult to understand interactions (e.g., max seq len is only applied at the sentence level).
> > > 
> > > 
> > > > Text cleaning and max sequence length must be applied before any other process (sentence segmentation, etc.)
> > > > Otherwise it causes issues with preproc functions (segmented html which is not cleaned by html_text) and leads to difficult to understand interactions (e.g., max seq len is only applied at the sentence level).
> > > 
> > > 
> > > @dvsrepo Setting a sequence limit before sentence segmentation could bring unexpected results
> > 
> > 
> > Only for the last sentence of the full text if it gets cut, but we really want to control the full text length before anything else as this will cause potentially very large texts to go downstream into tokenization, sentence splitting etc. Max sequence length should be used to avoid extremely large text chunks. Max nr of sentences can provide a more fine grained measure for size of text.
> 
> Can't we control the text size by combining max sequence length and max number of sentences?
> 
> If not, it's hard to see how we apply that maximum sequence length to record-type data.

Yes, let's try this!

PR #270, opened 2020-06-10
refactor(features): simplifiy features configuration

frascuchon:
The idea here is control configurable input features and, for now, keep only `word` and `character` features.

When new features will be required (or desired), we can progressively include them.


frascuchon:
> 👍
> I just resolved merge conflicts regarding the doc strings.

cool ! thanks

PR #281, opened 2020-06-12
sets prediction log path via command serve

dvsrepo:


src/biome/text/_model.py
```diff
@@ -68,7 +68,8 @@ class PipelineModel(allennlp.models.Model, allennlp.data.DatasetReader):
     def __init__(self, name: str, head: TaskHead):
         allennlp.models.Model.__init__(self, head.backbone.vocab)
         allennlp.data.DatasetReader.__init__(self, lazy=True)
-
+        
+        self.PREDICTION_FILE_NAME = 'predictions.json'
```
frascuchon, line 72: could you define at class level

class PipelineModel....
   PREDICTION_FILE_NAME  =  "predictions.json"

PR #291, opened 2020-06-15
Feat/add slot filling tutorial

dcfidalgo:
This PR adds the slot filling tutorial. You can find the tutorial [here](https://github.com/recognai/biome-text/blob/feat/add_slot_filling_tutorial/docs/docs/documentation/tutorials/Training_a_sequence_tagger_for_Slot_Filling.ipynb)

Apart from the tutorial there are several fixes:
- bug fix in the tokenclassifier
- i set `flatten` by default to False when reading json (cc @frascuchon )
- moved the basicConfig for the logging module to the init of the package (cc @frascuchon ). this one was driving me crazy to figure out how logging works when no handler is specified ...

I still have to improve the test, so it does not take ages, but tests a fair amount of functionality.

### DIFF
--- a/docs/docs/documentation/tutorials/Training_a_sequence_tagger_for_Slot_Filling.ipynb
+++ b/docs/docs/documentation/tutorials/Training_a_sequence_tagger_for_Slot_Filling.ipynb
@@ -1,0 +1,811 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Training a sequence tagger for Slot Filling"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<table>\n",
+    "  <td>\n",
+    "    <a target=\"_blank\" href=\"https://www.recogn.ai/biome-text/documentation/tutorials/Training_a_sequence_tagger_for_Slot_Filling.html\"><img src=\"https://www.recogn.ai/biome-text/assets/img/biome-isotype.svg\" width=32 />View on recogn.ai</a>\n",
+    "  </td>\n",
+    "  <td>\n",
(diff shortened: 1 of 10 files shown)


dcfidalgo:
Ok, this should be ready to review.

@frascuchon One important change that i forgot to mention is [this one](https://github.com/recognai/biome-text/pull/291/files#diff-1f508ce5914a3b884cfd60cb1ea07ceaL93).
When using a "pytorch_transformer" as encoder, the input dim to the encoder has to be divisible by the number of attention heads. This check gave me problems when setting the "embedding_dim" artificially to 1 before creating the vocab. I just removed the line, but i am not 100% sure about potential negative side effects.

dvsrepo:
Good job! My only comment would be if it's possible to define the pipeline programmatically instead of with the dict to show that possibility, once we merged now the spec renaming, would it be a lot of work?

dcfidalgo:
> Just one comment. Could we use the pypi package instead of git repo reference??

@frascuchon when i wrote the tutorial in the beginning i could not use the pypi version, since some essential commits were missing. The rc2 version should work now. Should i change it now or when we do the official release?

> Good job! My only comment would be if it's possible to define the pipeline programmatically instead of with the dict to show that possibility, once we merged now the spec renaming, would it be a lot of work?

@dvsrepo ok, i will work on it today.

frascuchon:
> Ok, this should be ready to review.
> 
> @frascuchon One important change that i forgot to mention is [this one](https://github.com/recognai/biome-text/pull/291/files#diff-1f508ce5914a3b884cfd60cb1ea07ceaL93).
> When using a "pytorch_transformer" as encoder, the input dim to the encoder has to be divisible by the number of attention heads. This check gave me problems when setting the "embedding_dim" artificially to 1 before creating the vocab. I just removed the line, but i am not 100% sure about potential negative side effects.

No problem. Go ahead

frascuchon:
> Good job! My only comment would be if it's possible to define the pipeline programmatically instead of with the dict to show that possibility, once we merged now the spec renaming, would it be a lot of work?

@dvsrepo  I will merge the PR. The suggested comments will be tackle in a separate PR

PR #293, opened 2020-06-17
Rename spec's to configuration for aligning concepts

dvsrepo:


dvsrepo:
@dcfidalgo the text classifier notebook test is failing, not sure why.

Otherwise this would be good to go after your comments

dvsrepo:
I have reviewed the examples roughly and did not find errors. I think we should in any case cleanup the examples folder.

dcfidalgo:
Looks good to me. I also think we should move the examples folder out of the repo. They are not used in the tests and also not mentioned in the docs, so they are floating in some limbo right now ...

@dvsrepo Just merged a PR fixing the tutorial test fail, if you rebase onto master it should pass.

frascuchon:
> Looks good to me. I also think we should move the examples folder out of the repo. They are not used in the tests and also not mentioned in the docs, so they are floating in some limbo right now ...
> 
> @dvsrepo Just merged a PR fixing the tutorial test fail, if you rebase onto master it should pass.

I'm quite using some examples when I need to test biome environment installation. If I can reproduce this with tutorials, (without wait for training long time), I'm okay with cleaning examples folder

dcfidalgo:
> I'm quite using some examples when I need to test biome environment installation. If I can reproduce this with tutorials, (without wait for training long time), I'm okay with cleaning examples folder

Then i would rather move them to the test folder, to keep the repo a bit better organized. Either just move the folder or really translate them to tests, if we do not want to use them in the CI, we could mark them as skip. In this way you can still use them manually.

dvsrepo:
Can we merge this?

dcfidalgo:
Just resolved a merge conflict. For my part this can go in.

PR #312, opened 2020-06-18
Update tutorials

dcfidalgo:
This PR addresses requested changes to the tutorials.

[text classifier](https://github.com/recognai/biome-text/blob/fix/tutorials/docs/docs/documentation/tutorials/Training_a_text_classifier.ipynb) (@leiyre, @ame):
 - rearrangement of icons
 - changed screenshot
 - TIP/WARNING -> Tip/Warning
Keep in mind that the rendering of the notebook here in github is not very indicative for the outcome in vuepress! 

[token classifier](https://github.com/recognai/biome-text/blob/fix/tutorials/docs/docs/documentation/tutorials/Training_a_sequence_tagger_for_Slot_Filling.ipynb) (@dvsrepo):
 - rearrangement of icons
 - create `PipelineConfig` programmatically
 - there is a slight inconsistency when creating a `PipelineConfig` from scratch: for example the `encoder` in the `CharFeatures` or the `feedforward` in the `TokenClassificationConfiguration` have to be specified as python dicts instead of `Seq2VecEncoderConfiguration` and `FeedForwardConfiguration`, respectively. I think we can not do much about this, but maybe you have an idea @frascuchon


dvsrepo:
looks good to me! Only two comments:

1. Regarding this:

>there is a slight inconsistency when creating a PipelineConfig from scratch: for example the encoder in the CharFeatures or the feedforward in the TokenClassificationConfiguration have to be specified as python dicts instead of Seq2VecEncoderConfiguration and FeedForwardConfiguration, respectively. I think we can not do much about this, but maybe you have an idea @frascuchon

I agree. I think best way to proceed is to open a separate enhancement issue, where we can directly point to the tutorial for describing the inconsistency and discuss if there's a solution.

2. If it's not a lot of work, I would remove the TrainerConfiguration part and just include a tip saying: if you want to tune the trainer or use cuda_device, you can pass a trainer = TrainerConfiguration(cuda_device...) to the train method.


Otherwise, we can merge.

dcfidalgo:
@dvsrepo addressed your comments. from my side this is ready to merge. Not sure why the CI is not kicking in ...

PR #313, opened 2020-06-18
Pass on a TokenizerConfiguration to the Tokenizer

dcfidalgo:
This PR tries to resolve a TODO:
```
    # TODO(dcfidalgo): Check if it is possible to pass in a `TokenizerConfiguration` to avoid duplication of arguments
```
A `Tokenizer` is now instantiated with just a `TokenizerConfiguration`. In this way we avoid the duplication of every single argument in the two classes which can lead to errors when trying to sync them. In fact, two arguments were already missing in the `TokenizerConfiguration` ...

### DIFF
--- a/src/biome/text/tokenizer.py
+++ b/src/biome/text/tokenizer.py
@@ -1,14 +1,15 @@
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
-from allennlp.common import FromParams
+from allennlp.common import FromParams, Params
 from allennlp.data import Token
 from allennlp.data.tokenizers import SentenceSplitter, SpacyTokenizer
 from allennlp.data.tokenizers.sentence_splitter import SpacySentenceSplitter
 
+from biome.text.configuration import TokenizerConfiguration
 from biome.text.text_cleaning import DefaultTextCleaning, TextCleaning
 
 
-class Tokenizer(FromParams):
+class Tokenizer:
     """Pre-processes and tokenizes the input text
     
     Transforms inputs (e.g., a text, a list of texts, etc.) into structures containing `allennlp.data.Token` objects.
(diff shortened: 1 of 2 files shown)


src/biome/text/tokenizer.py
```diff
@@ -20,61 +21,47 @@ class Tokenizer(FromParams):
...
-        self._fetch_spacy_model(lang)
+        self.lang = config.lang
+        self._fetch_spacy_model()
+
+        if config.segment_sentences is True:
+            self.segment_sentences = SpacySentenceSplitter(language=self.lang, rule_based=True)
+        elif config.segment_sentences is False:
+            self.segment_sentences = None
+        else:
+            self.segment_sentences = SentenceSplitter.from_params(Params(config.segment_sentences))
```
frascuchon, line 39: You must pass to allennlp `Params` deep copy to avoid to lost original configuration (required for `pipeline.config` property)

allennlp `from_params` read dictionary params using `pop` method and checks an empty dictionary at the end

src/biome/text/tokenizer.py
```diff
@@ -20,61 +21,47 @@ class Tokenizer(FromParams):
...
 
-        self.lang = lang
-        self.segment_sentences = segment_sentences
-        self.max_nr_of_sentences = max_nr_of_sentences
-        self.max_sequence_length = max_sequence_length
-        self.text_cleaning = text_cleaning or DefaultTextCleaning()
+        if config.text_cleaning is None:
+            self.text_cleaning = DefaultTextCleaning()
+        else:
+            self.text_cleaning = TextCleaning.from_params(Params(config.text_cleaning))
```
frascuchon, line 48: Same comment as above

src/biome/text/tokenizer.py
```diff
@@ -20,61 +21,47 @@ class Tokenizer(FromParams):
...
 
         self._base_tokenizer = SpacyTokenizer(
-            language=self.lang, start_tokens=start_tokens, end_tokens=end_tokens,
+            language=self.lang, start_tokens=config.start_tokens, end_tokens=config.end_tokens,
         )
 
-    def _fetch_spacy_model(self, lang):
+    def _fetch_spacy_model(self):
```
frascuchon, line 54: I prefer to keep language as required argument and decorate method as @staticmethod  if you don't mind

The following code could be read as good since we really don't known anything about the `self.lang` dependency in `_fetch_spacy_model`
```python
self._fetch_spacy_model()
self.lang = lang
```

src/biome/text/tokenizer.py
dcfidalgo: done, forgot about this behaviour!

src/biome/text/tokenizer.py
dcfidalgo: done

src/biome/text/tokenizer.py
dcfidalgo: done

PR #323, opened 2020-06-19
quick pass over API doc strings

dcfidalgo:
Another quick pass over the API doc strings.
The main additions are the TrainerConfiguration, WordFeatures and CharFeatures doc strings.
I am not sure if the arguments `cache_instances` and `in_memory_batches` are used at all at the moment, have to check this.

I also propose a slight change in the format: i would avoid specifying the type in the doc string, so this:
```python
def get_example(argument: int) -> str:
"""Gets an example name.

Parameters
----------
argument : int
    An argument

Returns
-------
example : str
    Name of the example
```
becomes this:

```python
def get_example(argument: int) -> str:
"""Gets an example name.

Parameters
----------
argument
    An argument

Returns
-------
example
    Name of the example
```
Since we consequently use type annotations the information is already in the signature of the method. Also the rendered html files look prettier without all the [colored boxes](https://gitlab.com/recognai-team/team/-/issues/122#note_362216441) in my opinion.

@dvsrepo @frascuchon what do you think? I would be willing to change the format for all present doc strings in a follow-up PR.

### DIFF
--- a/src/biome/text/about.py
+++ b/src/biome/text/about.py
@@ -5,20 +5,45 @@ __version__ = "1.0.0rc3"
 from typing import Optional
 
 
-def package_version(version: str):
-    """Generates a proper package version from git repository info"""
+def package_version(version: str) -> str:
+    """Generates a proper package version from the git repository info
+
+    Parameters
+    ----------
+    version
+
+    Returns
+    -------
+    package_version
+    """
 
     def get_commit_hash(repository: git.Git) -> str:
         """
-        Fetch current commit hash from configured git repository. The working folder should
-        be part of a already git repository
+        Fetch current commit hash from the configured git repository.
+        The working directory should already be part of a git repository.
 
+        Parameters
+        ----------
+        repository
+
+        Returns
+        -------
+        commit_hash
         """
         return repository.log("--pretty=format:'%h'", "-n 1").replace("'", "")
 
     def get_first_tag_for_commit(repository: git.Git, commit_hash: str) -> str:
-        """ Return tags related to current commit """
+        """Return tags related to current commit
 
+        Parameters
+        ----------
+        repository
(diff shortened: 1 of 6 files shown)


dvsrepo:
I agree with the format, I also notice is messy and more work with no gains as we have type annotations. Go ahead

frascuchon:
> I agree with the format, I also notice is messy and more work with no gains as we have type annotations. Go ahead

Just one thing. There is something cool typing arguments in docstring and is navigation. We can navigate to reference class by clicking argument type. If we remove this, we lost this behaviour.

frascuchon:
> I am not sure if the arguments `cache_instances` and `in_memory_batches`

As a said, there is an open research about lazy datasets vs in memory datasets and data bucketing that address directly to those parameters. We discussed at the time and the idea was keep as lazy instance (using only `batch_size` argument) and review when we're sure how to integrate with data bucketing.

But if you feel you can tackle the changes and drop both parameters,  please, go ahead

dvsrepo:
Yes, that behavior it's cool, the only issue is that I still don't know why sometimes it works and sometimes doesn't. Here is an example where it works in the signature but not in the params for the Trainer Config https://www.recogn.ai/biome-text/api/biome/text/configuration.html#trainer configuration. Generally it has issues with optionals, long or complex types. One thing that works fo sure is to mention simple types in the descriptions. Param typing sometimes introduces a lot of clutter and causes parsing errors with the use ` (see vocab configuration for example). So in general I think is a good idea to simplify that part an if we want we can include the name of the main biome class  in the description while describing the Param.

frascuchon:
So we close like that and we revise in a separated PR ? @dvsrepo

dvsrepo:
All good on my side yes

PR #324, opened 2020-06-22
Adds minimal task guides and removes other sections

dvsrepo:


dvsrepo:
I'd like to merge this, and then rewrite the Configuration guides. 

Configuration guides should avoid technical details and focus on the possible configuration and examples for each block (Pipeline, Datasource, etc).

We can live with just these two guides for now and will include more in the coming weeks.

PR #328, opened 2020-06-25
feat: quiet train and train logging file

frascuchon:
This PR 
- Adds a `quiet` train param for disable common logging info (keeping only warnings and errors)
- Logs to file training process logging messages

### DIFF
--- a/src/biome/text/__init__.py
+++ b/src/biome/text/__init__.py
@@ -1,8 +1,33 @@
+import logging
 import warnings
 from warnings import warn_explicit
-import logging
 
 import pkg_resources
+
+try:
+    import tqdm
+
+    class TqdmWrapper(tqdm.tqdm):
+        """
+        A tqdm wrapper for progress bar disable control
+
+        We must use this wrapper before any tqdm import (so, before any allennlp import). It's why we
+        must define at top package module level
+
+        We could discard this behaviour when this PR is merged: https://github.com/tqdm/tqdm/pull/950
+        and then just environment vars instead.
+
+        """
+
+        disable: bool = False
+
+        def __init__(self, *args, **kwargs):
+            kwargs["disable"] = TqdmWrapper.disable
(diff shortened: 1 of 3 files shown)


frascuchon:
> Thanks for tackling this! Two general comments:
> 
> 1. FileHandler
>    I would rather add this handler to the RootLogger (`logging.getLogger()`) to also capture logs that biome.text is producing, and just remove this handler again in the end.

I'm ok sending all logging, although at the moment we do not add much info to what allennlp logging generates. I will do de changes.


> 2. quiet
>    Even with `quiet=True` you will get the output of the `tqdm`s. The only way i figured out to disable them without changing allennlp is redirecting stdout and stderr. but this is super ugly. 

I found a programatic way to disable `tqdm`. I will test it to check if we really achieve a `quiet` training. 

> And even there it is not really necessary, just a bonus. I guess what i am saying is that we do not really need this feature for now, sorry for bringing it up in the call today!

Since we have a logging file, there is no reason to remove this functionality. In addition, being practically implemented it does not hurt. It could be interesting for background or concurrent executions that users want to launch, without having to do logging redirection hacks or things like that.

dcfidalgo:
> I found a programatic way to disable tqdm. I will test it to check if we really achieve a quiet training.

Nice!! If this works, than i will definitely use this feature in the hpo tutorial!

frascuchon:
New changes brings a fully quiet mode for trainings.

> > Thanks for tackling this! Two general comments:
> > 
> > 1. FileHandler
> >    I would rather add this handler to the RootLogger (`logging.getLogger()`) to also capture logs that biome.text is producing, and just remove this handler again in the end.
> 
> I'm ok sending all logging, although at the moment we do not add much info to what allennlp logging generates. I will do de changes.
> 

The way you must configure the logging system using python is quite special. So, I've finally made a selective logging configuration in training process: only allennlp and biome will be take into account. Setting just these two loggers, fetch a really cool quiet mode.

PR #329, opened 2020-06-26
sanitize input from ray tune for allennlp's from_params machinery

dcfidalgo:
This PR allows us to simplify the usage of ray tune's parameter choices.

For example: `tune.choice()` returns numpy types, and providing a `numpy.int32` as parameter for the hidden size fails, since it expects a python `int`.

### DIFF
--- a/src/biome/text/helpers.py
+++ b/src/biome/text/helpers.py
@@ -8,6 +8,7 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Type
 import torch
 import yaml
 from allennlp.data import TextFieldTensors
+from allennlp.common import util
 from elasticsearch import Elasticsearch
 
 from . import environment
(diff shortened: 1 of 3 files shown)


src/biome/text/helpers.py
```diff
@@ -231,3 +232,30 @@ def stringify(value: Any) -> Any:
     if isinstance(value, Iterable):
         return [stringify(v) for v in value]
     return str(value)
+
+
+def sanitize_for_params(x: Any) -> Any:
```
frascuchon, line 237: Could give a more general name and change uses of `allennlp.common.util.sanitize` by this new one? 
There are some code places where we're making data sanitize (maybe in another PR)

dcfidalgo: The purpose of the two methods is not exactly the same (one is to sanitize for serialization, this one is to sanitize for a more flexible from_params usage), but maybe they are close enough to use just one version. If you dont mind i would check this in another PR.

PR #330, opened 2020-07-01
Ray Tune tutorial

dcfidalgo:
This PR adds a HPO tutorial in which we use Ray Tune to perform a hyperparameter search. With this [link](https://colab.research.google.com/github/recognai/biome-text/blob/feat/ray_tune_demo/docs/docs/documentation/tutorials/Performing_a_hyperparameter_optimization_for_a_text_classifier.ipynb) you can have a look at it in Google Colab, i think it is the best way to review this.

There is still a section missing (Checking results), but maybe we can have a quick call tomorrow to have a look at this together.


dvsrepo:
Hi!

I have been testing this and it's looking good. Things missing from my point of view:

1.  Checking the results: briefly showing the best model with a more informative metric (e.g., validation_acc ), update the other tutorial to indicate the eval metric (as we do for slot filling), compare and mention if there's an improvement.

2. How we access (1) the best model with our typical workflow (Pipeline.from_pretrained(...)) or (2) we just have the best_config? If the latter is true, then how to access the best config and create a pipeline + trainer. Ideally we would aim for (1) but if that's not the case let's close this tutorial today with the current way of doing things.

dcfidalgo:
> 1. Checking the results: briefly showing the best model with a more informative metric (e.g., validation_acc ), update the other tutorial to indicate the eval metric (as we do for slot filling), compare and mention if there's an improvement.

Done

> 2. How we access (1) the best model with our typical workflow (Pipeline.from_pretrained(...)) or (2) we just have the best_config? If the latter is true, then how to access the best config and create a pipeline + trainer. Ideally we would aim for (1) but if that's not the case let's close this tutorial today with the current way of doing things.

I included option (1) in the *Checking the results* section.

Apart from that, i opened up an issue in ray tune about the logging of certain hyperparameters to tensorboard. If one wants to get insights about the importance of each hyperparameter (or get a feeling what kind of configurations work) tensorboard is by far the best way to do this. So this is still an open issue that could be solved if tune accepts the proposed fix.

dvsrepo:
Cool!

Just a last comment:

I would include some other choices to show what's possible, even dropout as we do for lr, like:

"type": tune.choice(['gru', 'lstm']),
 "num_layers": tune.choice([1, 2, 3]),
"hidden_size": tune.choice([16, 32, 64]),
"bidirectional": tune.choice([True, False]),

Then is good to go on my side!

dvsrepo:
All good on my side!

For the docs visualization:

- Leave some outputs that help to understand what's going on (e.g., the analysis dataframe?)
- I would switch to a title along the lines of **Hyperparameter optimization for a text classifier with Ray Tune**, I think is better for SEO and promoting this :-) Not sure about the exact phrasing ...for/of a text classifier ? If it sounds weird we could have "Hyperparameter optimization for NLP with Ray Tune" as the process would be highly similar for other type of models.

I promise it's my last comment :-)

dcfidalgo:
> Leave some outputs that help to understand what's going on (e.g., the analysis dataframe?)

Yes, i thought about what cell output to show, but honestly there is not much to show. Regarding the analysis dataframe, since i do not perfrom the hpo on my machine i do not really have a representative output ... I think i will simply add a screenshot of TensorBoard and the analysis DataFrame from Google Colabs.

> I would switch to a title along the lines of Hyperparameter optimization for a text classifier with Ray Tune, I think is better for SEO and promoting this :-) Not sure about the exact phrasing ...for/of a text classifier ? If it sounds weird we could have "Hyperparameter optimization for NLP with Ray Tune" as the process would be highly similar for other type of models.

Agreed, how about short and simple: **Hyperparameter optimization with Ray Tune** . Since this tutorial lives in the biome.text docs, it should be clear that it is about biome.text ...

dvsrepo:
Perfect!

PR #331, opened 2020-07-06
Add `Pipeline.predict_batch` method

dcfidalgo:
This PR adds a simple `Pipeline.predict_batch` method to process the input data as a batch.

I made a little comparison with the `Pipeline.predict()` method, and for a simple text classifier with 100 instances of a one token input, the `predict_batch` is 5-6 times faster. 

I would prefer to merge a simple `predict_batch()` version first, before adding further functionality (like batch_size, for example) in follow-up PRs.

As a side note: Depending on the input, using `predict` with the prediction cache can be much faster than `predict_batch`, just to keep this in mind.

### DIFF
--- a/src/biome/text/_model.py
+++ b/src/biome/text/_model.py
@@ -15,15 +15,13 @@ import numpy
 import torch
 from allennlp.common import Params
 from allennlp.common.util import sanitize
-from allennlp.data import DataLoader, Field, Instance, Token, Vocabulary
+from allennlp.data import Field, Instance, Token, Vocabulary
 from allennlp.data.fields import ListField, TextField
 from allennlp.models.archival import CONFIG_NAME
-from dask.dataframe import Series as DaskSeries
 
 from . import vocabulary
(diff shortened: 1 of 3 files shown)


src/biome/text/_model.py
```diff
@@ -277,6 +275,38 @@ def predict(self, *args, **kwargs) -> Dict[str, numpy.ndarray]:
...
+        """
+        if self.training:
+            warnings.warn(
+                "Train model enabled. "
+                "Disabling training mode automatically. You can manually disable it: "
+                "self.eval() or self.train(False)"
+            )
+            self.eval()
+
+        instances = [self.text_to_instance(**input_dict) for input_dict in input_dicts]
```
frascuchon, line 303: Just a comment but we must handle this at some point:

Since method `text_to_instance` could return an empty instance, we could have problems applying batch prediction here.

We had this problem in older versions. 

Maybe a metadata field keeping instance index in batch could resolve this problem but we must discuss about

PR #333, opened 2020-07-06
Feature: Train loggers integration

frascuchon:
This PR adds a new train component: The training logger

This component adds extra functionality to EpochCallback's (indeed, its implementation is based on callbacks) since define init and end training mechanisms that callbacks doesn't 

Also implements a `MlflowLogger`  using this new component for mlflow train logging as experiment runs.

### DIFF
--- a/src/biome/text/loggers.py
+++ b/src/biome/text/loggers.py
@@ -23,7 +23,6 @@ class BaseTrainLogger(EpochCallback):
 
         Parameters
         ----------
-
         pipeline:
             The training pipeline
         trainer_configuration:
--- a/src/biome/text/loggers.py
+++ b/src/biome/text/loggers.py
@@ -58,7 +58,6 @@ class BaseTrainLogger(EpochCallback):
             The current epoch
         metrics:
             The metrics related to current epoch
-
         """
         pass
 
--- a/src/biome/text/loggers.py
+++ b/src/biome/text/loggers.py
@@ -54,7 +54,6 @@ class BaseTrainLogger(EpochCallback):
 
         Parameters
         ----------
-
         epoch:
             The current epoch
         metrics:
--- a/src/biome/text/loggers.py
+++ b/src/biome/text/loggers.py
@@ -33,7 +33,6 @@ class BaseTrainLogger(EpochCallback):
             Validation dataset
         test:
             Test dataset
-
         """
         pass
 
--- a/src/biome/text/loggers.py
+++ b/src/biome/text/loggers.py
@@ -8,8 +8,8 @@ from biome.text.data import InstancesDataset
 from biome.text.training_results import TrainingResults
 
 
-class BaseTrainingLogger(EpochCallback):
-    """Base training logger for pipeline training"""
+class BaseTrainLogger(EpochCallback):
+    """Base train logger for pipeline training"""
 
     def init_train(
         self,
@@ -79,7 +79,7 @@ class BaseTrainingLogger(EpochCallback):
             self.log_epoch_metrics(epoch, metrics)
 
 
-class MlflowLogger(BaseTrainingLogger):
+class MlflowLogger(BaseTrainLogger):
     """A common mlflow logger for pipeline training"""
 
     def __init__(self, experiment_name: str = None, artifact_location: str = None):
--- a/src/biome/text/loggers.py
+++ b/src/biome/text/loggers.py
@@ -1,0 +1,132 @@
+from typing import Any, Dict, Optional
+
+from allennlp.training import EpochCallback, GradientDescentTrainer
+from mlflow.tracking import MlflowClient
+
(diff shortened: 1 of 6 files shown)


src/biome/text/pipeline.py
```diff
@@ -177,6 +179,8 @@ def train(
         epoch_callbacks:
             A list of callbacks that will be called at the end of every epoch, and at the start of
             training (with epoch = -1).
+        loggers:
+            A list of training loggers used in training. See `MlflowLogger` as example
```
dcfidalgo, line 183: Maybe we could explain this component a bit more in detail, something like:
```suggestion
            A list of loggers that execute a callback before the training, after each epoch, and at the end of the training (see `biome.text.logger.MlflowLogger`, for example)```

File: src/biome/text/loggers.py [hunk 2]
```diff
@@ -0,0 +1,131 @@
...
+        """
+        pass
+
+    def log_epoch_metrics(self, epoch: int, metrics: Dict[str, Any]):
+        """
+        Log epoch metrics
+
+        Parameters
+        ----------
+
```
dcfidalgo, line 60:
```suggestion
```

src/biome/text/loggers.py
```diff
@@ -0,0 +1,131 @@
...
+        training: InstancesDataset,
+        validation: Optional[InstancesDataset] = None,
+        test: Optional[InstancesDataset] = None,
+    ):
+        """
+        Init train logging
+
+        Parameters
+        ----------
+
```
dcfidalgo, line 26:
```suggestion
```

src/biome/text/loggers.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        pass
+
+    def end_train(self, results: TrainingResults):
+        """
+        End train logging
+
+        Parameters
+        ----------
+
```
dcfidalgo, line 47:
```suggestion
```

src/biome/text/loggers.py
```diff
@@ -0,0 +1,131 @@
...
+        results:
+            The training result set
+
```
dcfidalgo, line 50:
```suggestion
```

src/biome/text/loggers.py
```diff
@@ -0,0 +1,131 @@
...
+            The training pipeline
+        trainer_configuration:
+            The trainer configuration
+        training:
+            Training dataset
+        validation:
+            Validation dataset
+        test:
+            Test dataset
+
```
dcfidalgo, line 37:
```suggestion
```

src/biome/text/loggers.py
```diff
@@ -0,0 +1,131 @@
...
+        Log epoch metrics
+
+        Parameters
+        ----------
+
+        epoch:
+            The current epoch
+        metrics:
+            The metrics related to current epoch
+
```
dcfidalgo, line 65:
```suggestion
```

src/biome/text/loggers.py
```diff
@@ -0,0 +1,131 @@
...
+        pass
+
+    def __call__(
+        self,
+        trainer: GradientDescentTrainer,
+        metrics: Dict[str, Any],
+        epoch: int,
+        is_master: bool,
+    ):
+
```
dcfidalgo, line 76:
```suggestion
```

src/biome/text/pipeline.py
```diff
@@ -155,6 +156,7 @@ def train(
         test: Optional[Union[DataSource, InstancesDataset]] = None,
         extend_vocab: Optional[VocabularyConfiguration] = None,
         epoch_callbacks: List["allennlp.training.EpochCallback"] = None,
+        loggers: List[BaseTrainLogger] = None,
```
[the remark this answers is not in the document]
frascuchon: Ok for me. Just a comment about naming. 

Maybe `BaseTrainLogger` is a quite specific name for a genera component used at some training points, but I don't known other use cases about callback trainer more than logging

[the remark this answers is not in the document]
dvsrepo: It's ok on my side too.

src/biome/text/loggers.py
```diff
@@ -0,0 +1,125 @@
...
+    ):
+        if epoch >= 0:
+            self.log_epoch_metrics(epoch, metrics)
+
+
+class MlflowLogger(BaseTrainLogger):
+    """A common mlflow logger for pipeline training"""
+
+    def __init__(self, experiment_name: str = None, artifact_location: str = None):
+        self._client = MlflowClient()
```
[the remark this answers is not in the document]
frascuchon: Yes, tracking URI is always configurable by env var. For now, I prefer keep this behaviour for all mlflow dependent components, since could be common to all of them in the same project

PR #334, opened 2020-07-07
feat(pipeline): enable pipeline clone

frascuchon:
Adds fallback mechanism to pipeline clone when  deepcopy fails.

Note: This PR references pretrained model works

### DIFF
--- a/src/biome/text/pipeline.py
+++ b/src/biome/text/pipeline.py
@@ -202,18 +202,20 @@ class Pipeline:
             self.__configure_training_logging(output, quiet)
 
             # The original pipeline keeps unchanged
-            train_pipeline = self.__clone_pipeline()
+            train_pipeline = self.__make_copy()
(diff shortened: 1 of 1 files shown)


src/biome/text/pipeline.py
```diff
@@ -242,6 +241,20 @@ def train(
         finally:
             self.__restore_training_logging()
 
+    def __clone_pipeline(self):
+        """
+        Handles pipeline clone. Tries to apply a deepcopy as default and, if is not possible,
+        create pipelines from self instance type manually
+        """
+        try:
+            return copy.deepcopy(self)
```
dcfidalgo, line 250: Do we need this deepcopy? why not always create a Pipeline from scratch?

frascuchon: Not really, I will drop it.

PR #341, opened 2020-07-10
feat(pipeline): enable explain batch predictions

frascuchon:
This PR includes a new method that will apply explain over a prediction batch

### DIFF
--- a/src/biome/text/_model.py
+++ b/src/biome/text/_model.py
@@ -273,13 +273,15 @@ class PipelineModel(allennlp.models.Model):
         try:
             prediction = self.forward_on_instance(instance)
         except Exception as error:
-            raise WrongValueError(f"Failed to make a prediction for '{inputs}'") from error
+            raise WrongValueError(
+                f"Failed to make a prediction for '{inputs}'"
+            ) from error
         self.log_prediction(inputs, prediction)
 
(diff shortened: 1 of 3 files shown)


src/biome/text/_model.py
```diff
@@ -307,12 +309,48 @@ def predict_batch(
...
+            predictions = self.forward_on_instances(instances)
+        except Exception as error:
+            raise WrongValueError(
+                f"Failed to make predictions for '{input_dicts}'"
+            ) from error
+
+        explained_predictions = [
+            self.__build_explained_prediction(prediction, instance, n_steps)
+            for instance, prediction in zip(instances, predictions)
+        ]
```
dcfidalgo, line 350: I think this loop is not necessary. In the explain methods of the heads, we always embed the [instance in a batch](https://github.com/recognai/biome-text/blob/master/src/biome/text/modules/heads/classification/text_classification.py#L84). So the input to the `IntegratedGradients.attribute` is already a batch.
But let's leave this possible optimization to another PR (Maybe add an TODO here).

frascuchon: We will tackle in another PR. We need provide a common mechanism for prediction --> explain conversion if we accept explain prediction batches.

PR #342, opened 2020-07-10
[refactor] Avoid code duplication for predict/explain methods

dcfidalgo:
This PR is a small refactoring of the predict/explain methods:
- `explain` and `predict` use their corresponding `*_batch` methods
- A new helper function `_get_instances_and_predictions` avoids code duplication in the `predict_batch` and `explain_batch` methods.
- i copied the doc strings from the pipeline class, maybe there is a programmatic way to copy identical doc strings ?

### DIFF
--- a/src/biome/text/_model.py
+++ b/src/biome/text/_model.py
@@ -260,25 +260,9 @@ class PipelineModel(allennlp.models.Model):
 
     def predict(self, *args, **kwargs) -> Dict[str, numpy.ndarray]:
         """Make a single model prediction"""
-        if self.training:
-            warnings.warn(
-                "Train model enabled. "
-                "Disabling training mode automatically. You can manually disable it: "
-                "self.eval() or self.train(False)"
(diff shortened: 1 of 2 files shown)


src/biome/text/_model.py
```diff
@@ -312,74 +319,70 @@ def predict_batch(
             raise WrongValueError(
                 f"Failed to make predictions for '{input_dicts}'"
             ) from error
+
         for input_dict, prediction in zip(input_dicts, predictions):
             self.log_prediction(input_dict, prediction)
```
frascuchon, line 324: Just a comment. With this refactor all predictions (from predict and explain actions) will be logged.

It should be nice only apply prediction logging to predict actions. (But it's not a PR requirement at all)

dcfidalgo: Yes, i was not sure which behavior we prefer. i will change it, it's an easy fix.

PR #345, opened 2020-07-13
Feat/transformer

dcfidalgo:
I think this branch is finally ready to be merged. There was a tedious issue with how the transformers vocab is treated in allennlp.
With this PR we can start using the pretrained transformers in a simple way and check if we need a more fine-grained control over their usage.

@frascuchon In the end i add the transformers namespace to the `non_padded_namespaces`. I did not see this before, but by default the PretrainedTransformerIndexer gets the `tags` namespace, which is included in the default `non_padded_namespaces` ... I left a big comment in the `create_vocabulary` method explaining the hack.

@frascuchon i also moved the `create_vocabulary` to the `_BlankPipeline` class only.

Also added a simple integration test for the new feature.

### DIFF
--- a/src/biome/text/configuration.py
+++ b/src/biome/text/configuration.py
@@ -104,6 +104,7 @@ class FeaturesConfiguration(FromParams):
             embedder_cfg = configuration["word"]["embedder"]
             if "pretrained_file" in embedder_cfg:
                 embedder_cfg["pretrained_file"] = None
+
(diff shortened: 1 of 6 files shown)


dcfidalgo:
The new `__make_copy()` methods triggers calls to the random number generators -> this affects the training result of the integration test. Updated the results accordingly.

Issue #183, opened 2020-05-18
implement a learning rate finder


dcfidalgo:
With our new `config.json` we loose the ability to use the `allennlp find-lr` command: 
```terminal
(biome) macbook-pro-de-recognai:1 david$ allennlp find-lr -s find-lr --include-package=biome.text config.json
2020-05-18 16:42:20,680 - INFO - pytorch_pretrained_bert.modeling - Better speed can be achieved with apex installed from https://www.github.com/nvidia/apex .
2020-05-18 16:42:21,551 - INFO - pytorch_transformers.modeling_bert - Better speed can be achieved with apex installed from https://www.github.com/nvidia/apex .
2020-05-18 16:42:21,561 - INFO - pytorch_transformers.modeling_xlnet - Better speed can be achieved with apex installed from https://www.github.com/nvidia/apex .
2020-05-18 16:42:21,956 - INFO - allennlp.common.registrable - instantiating registered subclass relu of <class 'allennlp.nn.activations.Activation'>
2020-05-18 16:42:21,959 - INFO - allennlp.common.registrable - instantiating registered subclass relu of <class 'allennlp.nn.activations.Activation'>
2020-05-18 16:42:21,961 - INFO - allennlp.common.registrable - instantiating registered subclass relu of <class 'allennlp.nn.activations.Activation'>
2020-05-18 16:42:21,963 - INFO - allennlp.common.registrable - instantiating registered subclass relu of <class 'allennlp.nn.activations.Activation'>
Traceback (most recent call last):
  File "/Users/david/anaconda3/envs/biome/lib/python3.7/site-packages/allennlp/common/params.py", line 258, in pop
    value = self.params.pop(key)
KeyError: 'dataset_reader'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/david/anaconda3/envs/biome/bin/allennlp", line 8, in <module>
    sys.exit(run())

 File "/Users/david/anaconda3/envs/biome/lib/python3.7/site-packages/allennlp/run.py", line 18, in run
    main(prog="allennlp")
  File "/Users/david/anaconda3/envs/biome/lib/python3.7/site-packages/allennlp/commands/__init__.py", line 102, in main
    args.func(args)
  File "/Users/david/anaconda3/envs/biome/lib/python3.7/site-packages/allennlp/commands/find_learning_rate.py", line 127, in find_learning_rate_from_args
    force=args.force)
[...]

PR #346, opened 2020-07-15
[Feature] Implementing a `find_lr` method

dcfidalgo:
This PR introduces a new `Pipeline.find_lr()` method and closes issue #183 . The new method is based on AllenNLP's `search_learning_rate` function and just returns two lists: learning rates and losses, the user is responsible for plotting the data or apply some smoothing first.

Apart from the `find_lr()` method, this PR also includes some refactoring to reuse code, as well as exposing the `no_grad` option in the `TrainerConfiguration`.


src/biome/text/pipeline.py
```diff
@@ -154,6 +156,64 @@ def init_prediction_cache(self, max_size: int) -> None:
...
+            A configuration for finding the learning rate
+        training_data
+            The training data
+
+        Returns
+        -------
+        (learning_rates, losses)
+            Returns list of learning rates and corresponding losses.
+            Note: The losses are recorded before applying the corresponding learning rate
+        """
```
frascuchon, line 183: It would be nice include a minimal code snippet of how to use this method for learning rate selection (if possible)

dcfidalgo: Sorry, i totally forgot about this comment!
I planed to include a more detailed usage in one or two of our tutorials. There are a lot of blog posts/publications on how to select the learning rate once you made a learning rate scan, in my experience this learning rate scan gives you a rough estimate where to look for the learning rate, but to find the optimal one you still have to do a bit of HPO. I will include a link in the docstring pointing to one of those blog posts.

src/biome/text/pipeline.py
```diff
@@ -154,6 +156,66 @@ def init_prediction_cache(self, max_size: int) -> None:
...
+            A trainer configuration
+        find_lr_config
+            A configuration for finding the learning rate
+        training_data
+            The training data
+
+        Returns
+        -------
+        (learning_rates, losses)
+            Returns list of learning rates and corresponding losses.
```
dcfidalgo, line 183:
```suggestion
            Returns a list of learning rates and corresponding losses.
```

PR #357, opened 2020-08-10
Refactor: simplify spacy tokenization

frascuchon:
This PR tries to simplify integration with spacy at tokenization leve. By discarding all intermediate allennlp components, we gain in spacy dependency control and flow. This can be important for complex pipeline integrations (for data extraction) and some classification problems where tokenization affects to gold labels (Token classification)

### DIFF
--- a/src/biome/text/configuration.py
+++ b/src/biome/text/configuration.py
@@ -74,7 +74,11 @@ class FeaturesConfiguration(FromParams):
         char = CharFeatures(**char.as_dict(quiet=True)) if char else None
 
         transformers = params.pop("transformers", None)
-        transformers = TransformersFeatures(**transformers.as_dict(quiet=True)) if transformers else None
+        transformers = (
+            TransformersFeatures(**transformers.as_dict(quiet=True))
+            if transformers
+            else None
+        )
 
         params.assert_empty("FeaturesConfiguration")
         return cls(word=word, char=char, transformers=transformers)
@@ -152,7 +156,9 @@ class FeaturesConfiguration(FromParams):
         config_dict
         """
         configuration = {
-            spec.namespace: spec.config for spec in [self.word, self.char, self.transformers] if spec
+            spec.namespace: spec.config
+            for spec in [self.word, self.char, self.transformers]
+            if spec
         }
 
         return copy.deepcopy(configuration)
@@ -173,8 +179,11 @@ class TokenizerConfiguration(FromParams):
     text_cleaning
         A `TextCleaning` configuration with pre-processing rules for cleaning up and transforming raw input text.
     segment_sentences
-        Whether to segment input texts in to sentences using the default `SentenceSplitter` or
-        providing a specific configuration for a `SentenceSplitter`.
+        Whether to segment input texts in to sentences.
+    use_spacy_tokens
+        If True, tokenized token list will return spacy tokens instead allennlp tokens
+    remove_space_tokens
+        If True, all found space tokens will be removed from resultant token list.
     start_tokens
         A list of token strings to the sequence before tokenized input text.
     end_tokens
@@ -188,7 +197,9 @@ class TokenizerConfiguration(FromParams):
         max_sequence_length: int = None,
         max_nr_of_sentences: int = None,
         text_cleaning: Optional[Dict[str, Any]] = None,
-        segment_sentences: Union[bool, Dict[str, Any]] = False,
+        segment_sentences: bool = False,
+        use_spacy_tokens: bool = True,
+        remove_space_tokens: bool = True,
         start_tokens: Optional[List[str]] = None,
         end_tokens: Optional[List[str]] = None,
     ):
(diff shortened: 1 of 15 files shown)


src/biome/text/configuration.py
```diff
@@ -173,8 +179,11 @@ class TokenizerConfiguration(FromParams):
     text_cleaning
         A `TextCleaning` configuration with pre-processing rules for cleaning up and transforming raw input text.
     segment_sentences
-        Whether to segment input texts in to sentences using the default `SentenceSplitter` or
-        providing a specific configuration for a `SentenceSplitter`.
+        Whether to segment input texts in to sentences.
```
dcfidalgo, line 182:
```suggestion
        Whether to segment input texts into sentences.
```

src/biome/text/configuration.py
```diff
@@ -173,8 +179,11 @@ class TokenizerConfiguration(FromParams):
...
+    use_spacy_tokens
+        If True, tokenized token list will return spacy tokens instead allennlp tokens
```
dcfidalgo, line 184:
```suggestion
        If True, the tokenized token list contains spacy tokens instead of allennlp tokens
```

src/biome/text/configuration.py
```diff
@@ -173,8 +179,11 @@ class TokenizerConfiguration(FromParams):
...
+    remove_space_tokens
+        If True, all found space tokens will be removed from resultant token list.
```
dcfidalgo, line 186:
```suggestion
        If True, all found space tokens will be removed from the final token list.
```

src/biome/text/tokenizer.py
```diff
@@ -109,48 +99,50 @@ def _tokenize(self, text: str) -> List[Token]:
...
         tokens: `List[List[Token]]`
         """
-        sentences = [self.text_cleaning(text) for text in document]
+        paragraphs = [self.text_cleaning(text) for text in document]
         if self.segment_sentences:
-            sentences = [
-                sentence
-                for sentences in self.segment_sentences.batch_split_sentences(sentences)
-                for sentence in sentences
+            paragraphs = [
```
dcfidalgo, line 121: The name can be a bit confusing, `paragraphs` becomes a list of sentences here, right?

src/biome/text/tokenizer.py
```diff
@@ -109,48 +99,50 @@ def _tokenize(self, text: str) -> List[Token]:
...
+                sentence.string.strip()
+                for doc in self.__nlp__.pipe(paragraphs)
+                for sentence in doc.sents
             ]
+
         return [
-            self._tokenize(sentence)
-            for sentence in sentences[: self.max_nr_of_sentences]
+            self._tokenize(paragraph)
+            for paragraph in paragraphs[: self.max_nr_of_sentences]
```
dcfidalgo, line 129: The name `self.max_nr_of_sentences` can be a bit confusing here, since if `self.segment_sentences` is False, then `self.max_nr_of_sentences` will refer to the number of paragraphs.

frascuchon:
> Nice! Just a few inline comments.
> 
> I think the two inline comments about the `paragraphs` and `sentences` naming, root in the fact that i am not sure if it makes sense to convert a paragraphs hierarchy in the `tokenize_document()` method into a sentences hierarchy: by intuition i would have said that activating `self.segment_sentences` would add a new level to the hierarchy (document -> paragraph -> sentence), but i am not sure if we want or support this.

If I understand, you suggest no sentence splitting on document (list of text) tokenization?

src/biome/text/tokenizer.py
frascuchon: I think the problem here is the assumption that a document (list of text) is a list of sentences.

Document inputs can be generated by splitting a text by `\n`. In that case, the document may not be necessarily a list of sentences, but of paragraphs:

```python
doc = """
This is a sentence. This is another sentence.
One more sentence here
""".split("\n")

# In this case, doc will be:
doc == [ "This is a sentence. This is another sentence.", "One more sentence here" ] 

# ...and you still would like to split by sentence
[ "This is a sentence", "This is another sentence", "One more sentence here" ]
```

src/biome/text/tokenizer.py
frascuchon: Agree. Maybe limit a number of sentences make sense when `segment_sentences` is enabled. What do you think?

src/biome/text/tokenizer.py
dcfidalgo: If you want the hierarchy to be on sentence level, then you could just omit the first `.split("\n")` and pass in the whole `doc`.
Just saying that it is a bit counter intuitive that by activating `segment_sentences` you also through away the paragraph hierarchy.

dcfidalgo:
> If I understand, you suggest no sentence splitting on document (list of text) tokenization?

Yes, i think we provide the hierarchy implicitly with the list of text, so the `segment_sentences` option (the way it is implemented right now) only makes sense to me if the input is just a string.

src/biome/text/tokenizer.py
dcfidalgo: yes, like i wrote below, i am not sure if the `segment_sentences` option makes sense for a "list of text" input.

PR #360, opened 2020-08-28
Add WandBLogger

dcfidalgo:
This PR adds a simple `WandBLogger`.
If the `wandb` module can be imported and if a login is detected, all training runs will be logged automatically to your wandb account.

### DIFF
--- a/src/biome/text/loggers.py
+++ b/src/biome/text/loggers.py
@@ -159,16 +159,21 @@ class MlflowLogger(BaseTrainLogger):
     ):
         from pandas import json_normalize
 
-        # fmt: off
         for prefix, params_set in [
             ("pipeline", json_normalize(pipeline.config.as_dict())),
-            ("trainer", json_normalize(dataclasses.asdict(trainer_configuration)))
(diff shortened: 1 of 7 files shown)


src/biome/text/loggers.py
```diff
@@ -179,3 +188,67 @@ def end_train(self, results: TrainingResults):
...
+        self,
+        pipeline: "Pipeline",
+        trainer_configuration: "TrainerConfiguration",
+        training: InstancesDataset,
+        validation: Optional[InstancesDataset] = None,
+        test: Optional[InstancesDataset] = None,
+    ):
+        config = {
+            "pipeline": pipeline.config.as_dict(),
+            "trainer": dataclasses.asdict(trainer_configuration),
```
dvsrepo, line 225: Could we log here the `num_parameters`, it is an important factor for choosing/comparing models.

dvsrepo (changes requested):
Well done! Only a brief addition

src/biome/text/loggers.py
dcfidalgo: yes good point. Done

PR #368, opened 2020-09-17
Accept spans for the TokenClassification task

dcfidalgo:
This PR lets the TokenClassification head accept span labels as well as tag labels.

Still WIP, i think the current test fail is not related to the changes, have to test more tomorrow.

### DIFF
--- a/src/biome/text/modules/heads/token_classification.py
+++ b/src/biome/text/modules/heads/token_classification.py
@@ -1,7 +1,8 @@
+import logging
 from typing import Dict, List, Optional, Union, cast
 
 import torch
-from allennlp.common.checks import check_dimensions_match
+import spacy
 from allennlp.data import Instance, TextFieldTensors
 from allennlp.data.fields import SequenceLabelField, TextField
 from allennlp.modules import ConditionalRandomField, FeedForward, TimeDistributed
@@ -19,7 +20,21 @@ from .task_head import TaskHead, TaskName, TaskOutput
 
 
 class TokenClassification(TaskHead):
-    """Task head for token classification (NER)"""
+    """Task head for token classification (NER)
+
+    Parameters
+    ----------
+    backbone
+        The model backbone
+    labels
+        List of tag or span labels. Span labels get converted to tag labels internally.
+    label_encoding
+        The format of the tags. Supported encodings are: ['BIO', 'BIOUL']
+    top_k
+    dropout
+    feedforward
+    """
+    __LOGGER = logging.getLogger(__name__)
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


dcfidalgo:
@dvsrepo Ok, this should be ready to review. With this PR you can either provide "tag labels" (like bilou tags) or "span labels" in the format of a dictionary with "start", "end" and "label" keys. Span labels will be converted into tag labels internally, either bilou or bio tags. If there is a misalignment between the spans and the tokens, we issue a warning and discard the example for now.

src/biome/text/modules/heads/token_classification.py
```diff
@@ -31,6 +46,14 @@ def __init__(
         feedforward: Optional[FeedForwardConfiguration] = None,
     ) -> None:
         super(TokenClassification, self).__init__(backbone)
+
+        self._label_encoding = label_encoding
+
+        # Convert span labels to tag labels if necessary
+        # We just check if "O" is in the label list, a necessary tag for IOB/BIOUL schemes, an unlikely label for spans
+        if "O" not in labels:
```
dvsrepo, line 54: what happens if we have a label "O" for spans. Even if unlikely, can we test for this (both here and with a test in unit tests? can we warn you have "O" but not other IOB/BIOUL looking labels?

src/biome/text/modules/heads/token_classification.py
```diff
@@ -66,25 +89,74 @@ def __init__(
...
             self.backbone.vocab,
             tag_namespace=vocabulary.LABELS_NAMESPACE,
-            label_encoding=label_encoding,
+            label_encoding=self._label_encoding,
         )
 
         self.__all_metrics = [self.f1_metric]
         self.__all_metrics.extend(self.metrics.values())
 
+    def _convert_span_labels_to_tag_labels(self, labels: List[str]) -> List[str]:
```
dvsrepo, line 98: I'd like to move this and all methods related to entities conversion to a utils somewhere. The reason is I am working on a relation classifier, which will also receive stand-off annotations as a feature. Potentially other heads will benefit from having entities as features.

dvsrepo (changes requested):
Good job, just two comments to address and it's good to go!

src/biome/text/modules/heads/token_classification.py
dcfidalgo: > what happens if we have a label "O" for spans. Even if unlikely, can we test for this (both here and with a test in unit tests?

If you provide a label called "O" for spans, it will fail while training, because it will not find the "B-O" or "I-O" tags in the label vocabulary. Should i just test that this fails or do you want to support that use case? Maybe a specific parameter switch in the TaskHead, like `label_type = "spans"/"tags"`?

>  can we warn you have "O" but not other IOB/BIOUL looking labels?
Sure, i added a warning in that case.

src/biome/text/modules/heads/token_classification.py
dcfidalgo: Done

src/biome/text/modules/heads/token_classification.py
dvsrepo: Great! No need to support this use case or add `label_type`

dvsrepo (approved):
perfect!

PR #370, opened 2020-09-24
Baseline sentence/paragraph level relation classifier

dvsrepo:
This PR adds a basic relation classifier. Current caveats:

- We add an `entity_embedder` layer based on the `Embedding` allennlp layer. This works fine but we need to manually configure the `num_embeddings` param in the config, which somehow breaks current inference we have for embeddings (made possible by TextFieldEmbedders) and thus the extensibility of this embedding layer when further training.

- We manually concat this feature with the embedder output in the forward pass due to the same reason of not integrating this feature into the TextFieldEmbedder machinery.

I'm ok with this for now, but as discussed with @frascuchon this opens up the question of how to add other features to the `features` configuration (beyond current token text features). Both caveats will be avoided if we could:

- configure a `entities` feature in `features`.

- transform `entities` to tags in the featurizer, inject these tags into the tokenizer and add them to each token.

- have an indexer use this token._tag and the vocabulary namespace to create a new feature in the TextFieldEmbedder

- this will automatically concat this feature inside the embedder.

- this feature could be very general and not tied to entities, it could be other external token_level features such as POS tags, shape, len, etc. always making sure this features are token-level with 1:1 betwen tags and tokens (len(tokens) == len(tags))

- if we go a step further, one could define several token-level features (with some preprocessing of the training set) and use them in the `features` section. For example, I might do pos tagging, ner, etc. add this as feature to my dataset and then use them in the training process.  But this could of course create a mismatch between training and inference feature processing as we already know. In any case, some of this could be done internally leveraging spaCy and/or offsets (as done in this relation classifier) so we keep all feature processing inside the pipeline.

### DIFF
--- a/src/biome/text/helpers.py
+++ b/src/biome/text/helpers.py
@@ -3,11 +3,13 @@ import os
 import os.path
 import re
 from inspect import Parameter
-from typing import Any, Callable, Dict, Iterable, List, Tuple, Type
+from typing import Any, Callable, Dict, Iterable, List, Tuple, Type, Optional
 
+import spacy
 import yaml
 from allennlp.common import util
 from elasticsearch import Elasticsearch
+from spacy.tokens.doc import Doc
 
 from . import environment
 
@@ -273,7 +275,7 @@ def span_labels_to_tag_labels(
     return converted_labels
 
 
-def bioul_tags_to_bio_tags(tags: List[str]) -> List[str]:
+def _bioul_tags_to_bio_tags(tags: List[str]) -> List[str]:
     """Converts BIOUL tags to BIO tags
 
     Parameters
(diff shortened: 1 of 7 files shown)


dcfidalgo (approved):
Looks good to me, just a few inline comments.

As a general comment, i was wondering what the use case is of this model without the `entities_embedder`, since you made it optional? Without it, it's just a normal TextClassifier, right?

But i agree, i think treating the entities as a `feature` makes much more sense conceptually and maybe even technically, as you mentioned above. Also keep in mind that if one specifies an `encoder` in the pipeline config, the encoding is only applied to the `word`/`char`/`transformers` feature, since we concatenate the tensors after `self.backbone.forward(text, mask)`.

dvsrepo:
> Looks good to me, just a few inline comments.
> 
> As a general comment, i was wondering what the use case is of this model without the `entities_embedder`, since you made it optional? Without it, it's just a normal TextClassifier, right?

You are completely right, we should probably have it as mandatory, my idea was to compare the effect of adding this positional entity embeddings in my experiments, but I'm not sure its a needed baseline. Could you take this change over?

> But i agree, i think treating the entities as a `feature` makes much more sense conceptually and maybe even technically, as you mentioned above. Also keep in mind that if one specifies an `encoder` in the pipeline config, the encoding is only applied to the `word`/`char`/`transformers` feature, since we concatenate the tensors after `self.backbone.forward(text, mask)`.

Right! this another limitation of the current approach I forgot to mention. Yes, if possible it would make more sense to design this as a `feature`.

dcfidalgo:
I just made some adjustments i @dvsrepo could you take a look at the changes in this commit: https://github.com/recognai/biome-text/pull/370/commits/b29770e146ff91942f0de2368970e67c8fd38d91

I think it is actually important to treat the vocab for the `entities` input not as label or tag vocab. In a batch you will have some padding tokens for the text, so there should be a padding token for the entities as well. Furthermore, if we want to allow predictions with an untrained pipeline, we need either to provide all the entities in the config or we need a OOV token. Both tokens are not available when treating the entities vocab as label/tag vocab.

dvsrepo:
Perfect!

dcfidalgo:
@dvsrepo @frascuchon I will merge this PR tomorrow morning if there is no legal appeal by a judge until then ... and will open an issue to remind us that it is maybe better to treat the entities as a feature.

PR #372, opened 2020-09-25
Numeration of tutorials changed

ignacioct:
Numeration of the tutorials into the biome website should be changing to match the order of difficulty.


dcfidalgo (approved):
LGTM perfecto

PR #373, opened 2020-09-28
Numeration of tutorials on file names

ignacioct:
To correctly change the order in which tutorial show in the website


dcfidalgo:
Thanks @ignacioct I think having a number in the url is fine, for now we can leave it like this.

dcfidalgo (approved):

PR #374, opened 2020-09-28
Tutorials links fixed #274

### LINKED ISSUE #274: Make links absolute in tutorials
**Is your feature request related to a problem? Please describe.**
When following a tutorial in Google Colab, the links do not work.

**Describe the solution you'd like**
Make links absolute so that they work in Google Colab

ignacioct:
All links on the tutorial has been changed to absolute links to the biome website. All should be usable from all platforms.


dcfidalgo:
@ignacioct Thanks for this fix, very much appreciated!
@leiyre @amelie With this change all links in the tutorials will be "external links" that right now have a little symbol to their right and open in a new tab. I think this is a trade-off worth making, to avoid broken links when actually doing the tutorial in a notebook or on Colab. Just wanted to let you know.

PR #376, opened 2020-09-29
Fix/tutorial links

ignacioct:


dcfidalgo:
> @ignacioct Thanks for this fix, very much appreciated!
@leiyre @Amelie With this change all links in the tutorials will be "external links" that right now have a little symbol to their right and open in a new tab. I think this is a trade-off worth making, to avoid broken links when actually doing the tutorial in a notebook or on Colab. Just wanted to let you know.

> I will merge this PR later this day if there is no objection.

PR #382, opened 2020-10-02
Rename "annotation" to "prediction" in explore output

dcfidalgo:
This PR refactors a bit the explore output: The confusing "annotation" key gets a more descriptive name: "prediction" since it holds the output of the `Pipeline.predict()` command.

I tried to find all usages in the UI code, but not sure if i caught all of them. @frascuchon @leiyre is there some test suite i can run for the UI? I ran our first tutorial with the explore part and it looked fine.


frascuchon:
This changes will break already indexed explores/feedbacks. To minimize problems, i would like keep temporally both key names in UI pages until we are sure we can drop `annotations`

dcfidalgo:
> This changes will break already indexed explores/feedbacks. To minimize problems, i would like keep temporally both key names in UI pages until we are sure we can drop annotations

@frascuchon Is there an easy way the UI could accept both keys? My JS skills are very very modest ...

frascuchon (approved):

leiyre (approved):

PR #384, opened 2020-10-05
Replace call to non-existent _default_ds_mapping attribute

dcfidalgo:
This PR fixes a bug when using a `DataSource` with no mapping in the `Pipeline.explore` method.

When following our first tutorial line by line, you do not stumble on this bug, since we add a mapping to the data sources when calling the `Pipeline.train` method. But when you skip the training, create a DataSource without mapping and feed this data source to the explore method, the bug appears.

### DIFF
--- a/src/biome/text/pipeline.py
+++ b/src/biome/text/pipeline.py
@@ -427,10 +427,7 @@ class Pipeline:
         A torch Dataset containing the instances collection
 
         """
-        mapping = {k: k for k in self.inputs + [self.output] if k}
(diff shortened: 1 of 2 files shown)


frascuchon (commented):
Is easy include a minimal unit test checking the bug fixing ?

dcfidalgo:
> Is easy include a minimal unit test checking the bug fixing ?

Done

frascuchon (approved):

PR #386, opened 2020-10-05
Output entities for token classification

dcfidalgo:
This PR introduces entities in the output of the `TokenClassification` if the input text is a `str`.

For now the format of the entities is: `{"start": int, "end": int, "label": str}`. This format is also used as "label input" during training.

### DIFF
--- a/src/biome/text/modules/heads/token_classification.py
+++ b/src/biome/text/modules/heads/token_classification.py
@@ -1,9 +1,9 @@
 import logging
-from typing import Dict, List, Optional, Union, cast
+from typing import Dict, List, Optional, Union, cast, Tuple
 
 import torch
 from allennlp.data import Instance, TextFieldTensors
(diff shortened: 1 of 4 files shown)


src/biome/text/modules/heads/token_classification.py
```diff
@@ -158,30 +163,14 @@ def featurize(
...
-        )
-
-        del output.k_tags
-        return output
-
     def forward(  # type: ignore
-        self, text: TextFieldTensors, labels: torch.IntTensor = None
+        self,
+        text: TextFieldTensors,
+        raw_text: Union[str, List[str]],
```
frascuchon, line 172: If I undestand, `raw_text` contains texts/tokens related to input batch, so `Union[List[str], List[List[str]]]` ?

frascuchon (approved):
just a minimal comment

PR #388, opened 2020-10-07
Introduction of a TransformersTokenizer

dcfidalgo:
This PR introduces a `TransformersTokenizer` that enables us to directly process subword tokens usually used in the pretrained transformer models.

Until now, we bypassed a direct processing of subword tokens using AllenNLP mismatched indexer/embedder. With these changes here, if you only use the `transformers` feature, by default it will use the new `TransformersTokenizer` returning word pieces, and it will automatically use the matched indexer/embedder for the transformers feature.

This also allows us to use special tokens like for example the `CLS` token (used in many BERT-like models) for classification. You can check out the tests to see how to use the `bert_pooler` as pooler in our ClassificationHead.

The idea is that @ignacioct makes some comparisons between only subword tokens + bert_pooler, word tokens + gru pooler, word tokens with scalar mix + gru pooler ... etc with the new tutorial data. I think it will be interesting to see how they compare with each other.

(A lot of the additions are due to the tests, so don't be scared of reviewing the 183 lines of changes ...)

### DIFF
--- a/src/biome/text/configuration.py
+++ b/src/biome/text/configuration.py
@@ -225,8 +225,6 @@ class TokenizerConfiguration(FromParams):
         return all(a == b for a, b in zip(vars(self), vars(other)))
 
 
-
-
 class PipelineConfiguration(FromParams):
     """Creates a `Pipeline` configuration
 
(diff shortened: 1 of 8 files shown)


frascuchon:
@dcfidalgo using transformers for token classification will have any impact with this new tokenization approach?

dcfidalgo:
@frascuchon Yes, token classification becomes a bit more complicated when dealing with word pieces and our `TokenClassification` can still not deal with them. So if you want to use the transformers feature with our `TokenClassification` head, you have to use it with word tokens as we are doing now.

I will add a ConfigurationError when trying to use the `TokenClassification` head together with word pieces.

dcfidalgo:
@frascuchon Added a `NotImplementedError` when trying to use the `TokenClassification` head together with the `TransformersTokenizer`.

PR #393, opened 2020-10-08
feat: hpo components

frascuchon:
This PR simplify ray.tune hpo integration

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,7 @@ if __name__ == "__main__":
             "captum~=0.2.0",
             "ipywidgets~=7.5.1",
             "mlflow~=1.9.0",
-            "ray>=0.8.6,<0.9.0"
+            "ray~=1.0.0"
         ],
         extras_require={
             "testing": [
--- a/setup.py
+++ b/setup.py
@@ -95,6 +95,7 @@ if __name__ == "__main__":
             "captum~=0.2.0",
             "ipywidgets~=7.5.1",
             "mlflow~=1.9.0",
+            "ray>=0.8.6,<0.9.0"
         ],
         extras_require={
             "testing": [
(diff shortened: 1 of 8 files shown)


src/biome/text/features.py
```diff
@@ -61,7 +61,10 @@ def config(self) -> Dict:
 
     def to_json(self) -> Dict:
         """Returns the config as dict for the serialized json config file"""
-        return self.to_dict()
+        data = vars(self)
+        data.update(data.pop("extra_params", {}))
+
+        return data
 
     def to_dict(self) -> Dict:
```
dcfidalgo, line 69: This is the same as `to_json`, not sure why we have both methods really ... But i would either let `to_json` return `to_dict` or `to_dict` return `to_json`, not implementing the logic twice.

src/biome/text/features.py
```diff
@@ -140,7 +143,10 @@ def config(self) -> Dict:
 
     def to_json(self):
         """Returns the config as dict for the serialized json config file"""
-        return self.to_dict()
+        data = vars(self)
+        data.update(data.pop("extra_params", {}))
+
+        return data
 
     def to_dict(self):
```
dcfidalgo, line 151: See comment above

setup.py
```diff
@@ -95,6 +95,7 @@ def about_info(package: str):
             "captum~=0.2.0",
             "ipywidgets~=7.5.1",
             "mlflow~=1.9.0",
+            "ray>=0.8.6,<0.9.0"
```
dcfidalgo, line 98: they already released 1.0, can we use that?

dcfidalgo (commented):
Thanks! Just a few inline comments.

src/biome/text/features.py
frascuchon: only keep `to_json` for allenlp `sanitize` integration

src/biome/text/features.py
frascuchon: same as before

setup.py
frascuchon: done

dcfidalgo (approved):

PR #396, opened 2020-10-13
Refactor: separate explore from pipeline

frascuchon:
This PR create specific functions for data exploration interactions (`create explore` and  `show explore`), removing  `explore` method from `Pipeline` class


dcfidalgo (approved):
Nice! Yes, the `_explore` helper function has to be rewritten, also in the light of maybe replacing the DataSource with Huggingface's Datasets.

PR #397, opened 2020-10-13
Accept huggingface's datasets as input

dcfidalgo:
With this PR on can use [Huggingface's datasets](https://huggingface.co/docs/datasets/index.html) as input.

I tried to do some profiling to compare the two cases: using a `datasets.dataset` and our current `DataSource`:

**datasource**
```terminal
Line #    Mem usage    Increment   Line Contents
================================================
     2     53.8 MiB     53.8 MiB   def my_func():
     3    419.5 MiB    365.7 MiB       from biome.text import Pipeline
     4    419.5 MiB      0.0 MiB       from biome.text.configuration import VocabularyConfiguration, TrainerConfiguration
     5    419.5 MiB      0.0 MiB       from biome.text.data import DataSource
     6    419.5 MiB      0.0 MiB       from biome.text import loggers
     7    423.3 MiB      3.8 MiB       from datasets import load_dataset
     8                                 
     9    423.3 MiB      0.0 MiB       loggers._HAS_WANDB = False
    10                                 
    11                             #     ds = load_dataset("json", data_files="./arxiv-dataset-train_small.json", split="train")
    12                             #     ds.cleanup_cache_files()
    13                             #     ds = ds.map(lambda x: {"text": x["abstract"], "label": x["categories"]})
    14                             #     labels = list(set(ds["label"]))
    15                                 
    16    425.6 MiB      2.3 MiB       ds = DataSource("/home/david/recognai/biome/hf_datasets/arxiv-dataset-train_small.json", mapping={"text": "abstract", "label": "categories"})
    17    426.1 MiB      0.6 MiB       labels = list(set(ds.to_dataframe()["categories"]))
    18                                 
    19    426.1 MiB      0.0 MiB       pl = Pipeline.from_config({
    20    426.1 MiB      0.0 MiB           "name": "datasets_test",
    21                                     "features": {
    22    426.1 MiB      0.0 MiB               "word": {"embedding_dim": 16, "lowercase_tokens": True},
    23                                     },
    24                                     "head": {
    25    426.1 MiB      0.0 MiB               "type": "TextClassification",
    26    536.4 MiB    110.2 MiB               "labels": labels,
    27                                     },
    28                                 })
    29                                 
    30    536.4 MiB      0.0 MiB       vocab_config = VocabularyConfiguration(sources=[ds])
    31   1273.3 MiB    736.9 MiB       pl.create_vocabulary(vocab_config)
    32                                 
[...]

PR #398, opened 2020-10-14
Adding a Dataset class based on HuggingFace's datasets.Dataset

dcfidalgo:
This PR adds a new class `Dataset` intended to replace the current `DataSource`. It is basically a thin wrapper around HuggingFace's `datasets.Dataset` class enhanced by the for us crucial `to_instances` method, and a few convenient methods. See the advantages of this approach in #397 . 

In most cases i think we will use the `from_json` or `from_csv` class methods, but you have access to the whole `load_dataset` method, which is also used to create [HF datasets](https://huggingface.co/docs/datasets/loading_datasets.html). This gives us instant access to a lot of NLP datasets!

Indexing and iterating works directly with this class, but you have all the `datasets` utilities accessible via the underlying `dataset` attribute:

```python
from biome.text import Dataset

ds = Dataset.from_json(paths="my_json_file.jsonl")

# a nice repr for the dataset
print(ds)

# First 10 entries of the dataset
ds[:10]

# iterating over the dataset
for row in ds:
    print(row)

# advanced utilities for working with your dataset 
# (see https://huggingface.co/docs/datasets/master/processing.html)
ds.dataset = ds.dataset.map(...)
```

Further PRs will address following todos:
- make it work with our explore method
- implement a working cache mechanism for the instances
- Replace the `DataSource` class in the whole library

### DIFF
--- a/src/biome/text/dataset.py
+++ b/src/biome/text/dataset.py
@@ -81,6 +81,40 @@ class Dataset:
         """
         return cls.load_dataset("csv", data_files=paths, split="train", **kwargs)
 
+    @classmethod
+    def from_pandas(cls, df: "pandas.DataFrame", **kwargs):
+        """Convenient method to create a Dataset from a `pandas.DataFrame`
+
+        Parameters
+        ----------
+        df
+            The data frame
+        **kwargs
+            Passed on to `datasets.Dataset.from_pandas` method
+
+        Returns
+        -------
+        dataset
+        """
+        return cls(datasets.Dataset.from_pandas(df=df, **kwargs))
+
+    @classmethod
+    def from_dict(cls, mapping: dict, **kwargs):
+        """Convenient method to create a Dataset from a python dictionary
+
+        Parameters
+        ----------
+        mapping
+            A mapping of strings to arrays or python lists.
+        **kwargs
+            Passed on to `datasets.Dataset.from_dict` method
+
+        Returns
+        -------
+        dataset
+        """
+        return cls(datasets.Dataset.from_dict(mapping=mapping, **kwargs))
+
     def to_instances(self, pipeline: "biome.text.Pipeline", lazy=True) -> InstancesDataset:
         """Convert input to instances for the pipeline
 
--- a/src/biome/text/dataset.py
+++ b/src/biome/text/dataset.py
@@ -6,8 +6,6 @@ from typing import Union, Dict, Iterable, List
 import datasets
 from allennlp.data import AllennlpDataset, AllennlpLazyDataset, Instance
 
-from biome.text import Pipeline
-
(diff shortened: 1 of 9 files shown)


src/biome/text/dataset.py
```diff
@@ -0,0 +1,159 @@
...
+import pickle
+from typing import Union, Dict, Iterable, List
+
+import datasets
+from allennlp.data import AllennlpDataset, AllennlpLazyDataset, Instance
+
+InstancesDataset = Union[AllennlpDataset, AllennlpLazyDataset]
+
+
+class Dataset:
```
frascuchon, line 12: In terms of backward compatibility for data readers. Is there dataset readers for `parquet` or  `excel` formats? Could we create our own data readers easily ?

src/biome/text/dataset.py
```diff
@@ -0,0 +1,159 @@
...
+        Parameters
+        ----------
+        pipeline
+            The pipeline for which to create the instances.
+        lazy
+            If true, instanes are lazily read from disk, otherwise they are kept in memory.
+        """
+        self._LOGGER.info("Creating instances ...")
+
+        input_columns = [col for col in pipeline.inputs + [pipeline.output] if col]
```
frascuchon, line 96: using new datasets, you must adapt dataset format to pipeline inputs/outputs isn't it?

frascuchon (commented):
Very very nice!!!

just a couple of questions before merge.

src/biome/text/dataset.py
dcfidalgo: No there is no builtin reader for parquet or excel. It is possible to implement a custom reader but i am not sure if this is straight forward, i haven't had a detailed look at it yet: https://huggingface.co/docs/datasets/master/add_dataset.html

Since we used parquet mainly internally, i think we could switch to their format (arrow format) via `Dataset.save_to_disk`/`Dataset.load_from_disk`. If we want to support excel explicitly, i think we have to implement it ourselves.

src/biome/text/dataset.py
dcfidalgo: If you mean the mapping, yes, one has to make sure that the dataset contains the "right" column names. But this can be easily done with the Huggingface `Dataset` methods, for example:
```python
ds.dataset.rename_column_("old_col_name", "new_col_name")
# or combining stuff
ds.datasest = ds.dataset.map(lambda x: {"new_col_name": x["old_col_name"] + x["old_col_name2"]})
```

src/biome/text/dataset.py
frascuchon: mkay !

src/biome/text/dataset.py
frascuchon: cool !

frascuchon (approved):

PR #407, opened 2020-10-19
Make explore work with the new `Dataset`

dcfidalgo:
This PR tries to make the new `Dataset` class work together with our explore command.
It is still a draft that is failing, i did not manage to get the interaction with ES right.

@frascuchon i think i need your help here. When calling `dataset.add_elasticsearch_index` it tells me that the index already exists (`elasticsearch.exceptions.RequestError: RequestError(400, 'resource_already_exists_exception',`), because we create an empty index in a first step. So i am not sure if we can use this `Dataset` method ... Not sure how to proceed. I added a test where you can check out the error yourself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
-services:
-  - elasticsearch
+#services:
+#  - elasticsearch
 
 python:
   - "3.7"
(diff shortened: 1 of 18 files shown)


frascuchon:
@dcfidalgo Ok. I'll take care of this... thanks!

frascuchon (approved):
@dcfidalgo Elasticsearch index generation is working now. It's ok for me!

frascuchon:
Just a comment. I've found some segmentation fault problems with `test_explore` and some local warnings with threading. I think is something related to local ray cluster but we must investigate deeper. Just to keep in mind.

[Here](https://travis-ci.com/github/recognai/biome-text/builds/191394489) a build error with segmentation fault

PR #408, opened 2020-10-19
fix(setup): upgrade required pip version

frascuchon:
This PR upgrade required pip version. Since we're using feature `--use-feature` in makefile directive and that feature is included since pip 20.2 version


dcfidalgo (approved):
Thanks a lot! Forgot to update this when i added the `--use-feature` option ... :0

PR #409, opened 2020-10-19
refactor(token_classifier): config only accepts span labels

frascuchon:
This PR simplifies token classifiers configuration. Only accepts labels values a span values, and internally converts into configured labels encoding.

### DIFF
--- a/docs/docs/documentation/tutorials/2-Training_a_sequence_tagger_for_Slot_Filling.ipynb
+++ b/docs/docs/documentation/tutorials/2-Training_a_sequence_tagger_for_Slot_Filling.ipynb
@@ -40,7 +40,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -71,7 +71,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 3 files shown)


dcfidalgo (commented):
I would vote for keeping the flexibility of accepting pre-tokenized input together with their tags.

Most of the canonical data sets for NER (like ConNLL or the ones present in huggingface's datasets) are provided in form of pre-tokenized words + NER tags. So i don't think we should close the door for these datasets.

Also from the Cantemist experience i would say that if extensive data preprocessing is necessary, passing on pretokenized text to the model is more convenient. For Cantemist we did a lot of checks tokenizing the data in the data preprocessing, to find the tokens that messed up the alignment or caused troubles for the transformer tokenizers. If the model only had accepted text strings with spans, than we would have had to transform everything back to text and spans after removing/modifying the tokens. 

But if we want to accept only text strings + spans, i would also limit the predicted output to spans.

dcfidalgo:
Ah sorry, i totally misunderstood your changes! The only difference you propose is basically at the config level for the `TaskHead` where you specify the `labels`, so:
```python
    pipeline_dict = {
        "name": "biome-bimpm",
        "head": {
            "type": "TokenClassification",
            "labels": ["B-NER", "I-NER", "O"],
            "label_encoding": "BIO",
        },
    }
```
->
```python
    pipeline_dict = {
        "name": "biome-bimpm",
        "head": {
            "type": "TokenClassification",
            "labels": ["NER"],
            "label_encoding": "BIO",
        },
    }
```
even in case of pre-tokenized data, right?

That's fine with me, although in practice most of the times i do this:
```python
    pipeline_dict = {
        "name": "biome-bimpm",
        "head": {
            "type": "TokenClassification",
            "labels": list(set(dataset["label"])),
            "label_encoding": "BIO",
        },
    }
```
No strong opinion, so LGTM!

dcfidalgo (approved):

dcfidalgo:
As a reminder, if we want to merge this we also should update the tutorial 2!

frascuchon:
> Ah sorry, i totally misunderstood your changes! The only difference you propose is basically at the config level for the `TaskHead` where you specify the `labels`, so:
> 
> ```python
>     pipeline_dict = {
>         "name": "biome-bimpm",
>         "head": {
>             "type": "TokenClassification",
>             "labels": ["B-NER", "I-NER", "O"],
>             "label_encoding": "BIO",
>         },
>     }
> ```
> 
> ->
> 
> ```python
>     pipeline_dict = {
>         "name": "biome-bimpm",
>         "head": {
>             "type": "TokenClassification",
>             "labels": ["NER"],
>             "label_encoding": "BIO",
>         },
>     }
> ```
> 
> even in case of pre-tokenized data, right?
> 
> That's fine with me, although in practice most of the times i do this:
> 
> ```python
>     pipeline_dict = {
>         "name": "biome-bimpm",
>         "head": {
>             "type": "TokenClassification",
>             "labels": list(set(dataset["label"])),
>             "label_encoding": "BIO",
>         },
>     }
> ```
> 
> No strong opinion, so LGTM!

Yes, just at config level. Using this config approach, we could try to fix (or warn) some pre-tokenized training data if some encoded label is wrong.

frascuchon:
> As a reminder, if we want to merge this we also should update the tutorial 2!

Done

Issue #406, opened 2020-10-16
Add a `Pipeline.evaluate` method


dcfidalgo:
TODO: Implement a `Pipeline.evaluate` method that takes as input a `Dataset` and returns some metrics. We should expose this method to the CLI as well.

PR #410, opened 2020-10-19
[Draft] Add a `Pipeline.evaluate` method

dcfidalgo:
This PR adds an `evaluate` method to our `Pipeline` class addressing issue #406 . I left some todos, since i want to have a quick discussion first, before considering this for a merge.

Right now the `Pipeline` class has the `predict`/`explain`/`evaluate` methods that only really make sense for the `_PreTrainedPipeline` class. To make a meaningful prediction you have following flow:

- `pl = Pipeline.from_config()` -> `pl.train()` -> `pl = Pipeline.from_pretrained()` -> `pl.predict` /`pl.evaluate`

The reason is that we do not modify the weights "in-place", but create a copy of the pipeline when we train.
The **advantages** of an "in-place" modification would be:
- allows for a straight forward flow: `pl.from_config` -> `pl.train` -> `pl.predict`/`pl.evaluate` (right now, this flow will "work", but with an unexpected result);
- less memory footprint, no need to keep two models in memory when training;

**Disadvantages** would be (maybe i am missing some!):
- consecutive trainings from scratch would need an intermediate reset step: `pl.from_config()` -> `pl.train` -> `pl.reset` -> `pl.train`

In summary i vote for either: 
- implementing some sort of `reset` method that allows to reset a model to its initial state after a training, and modify the weights in-place,
- or moving the inference methods to the `_PreTrainedPipeline` class, although this will break the possibility of calling the `predict` method to test a pipeline configuration.

@frascuchon what do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
-services:
-  - elasticsearch
+#services:
+#  - elasticsearch
 
 python:
   - "3.7"
@@ -22,11 +22,24 @@ cache:
 before_install:
   - sudo apt-get update
   - sudo apt-get install nodejs -q -y
+  # Elasticsearch 7.x installation
+  - sudo apt purge elasticsearch
+  - sudo apt-get install apt-transport-https
+  - echo "deb https://artifacts.elastic.co/packages/7.x/apt stable main" | sudo tee /etc/apt/sources.list.d/elastic-7.x.list
+  - sudo apt-get update && sudo apt-get install elasticsearch=7.4.2 -q -y --allow-unauthenticated
+  - sudo sudo systemctl enable elasticsearch
+  - sudo systemctl start elasticsearch
+  - sudo systemctl status elasticsearch
+
 
 install:
   - pip install -U pip
   - pip install -U six
   - make dev
+
+before_script:
+  # Show elasticsearch version info
+  - curl -XGET http://localhost:9200
 
 script:
   - make dist docs
--- a/docs/docs/documentation/tutorials/2-Training_a_sequence_tagger_for_Slot_Filling.ipynb
+++ b/docs/docs/documentation/tutorials/2-Training_a_sequence_tagger_for_Slot_Filling.ipynb
@@ -40,7 +40,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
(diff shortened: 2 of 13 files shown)


dcfidalgo:
Moving the discussion from #411 to here.

In #411 @frascuchon suggested a very nice idea. We could have a context manager that restores the model state on exit.

So if you want to do consecutive trainings from scratch you could do something like this:
```python
with pl.restore_model():
    pl.train()
```

frascuchon:
@dcfidalgo It's this PR still in draft ?

dcfidalgo:
> @dcfidalgo It's this PR still in draft ?

I think it should work the way it is implemented in this PR. But maybe i would like to discuss the issues raised above first, before merging this.

frascuchon:
> In summary i vote for either:
>
> implementing some sort of reset method that allows to reset a model to its initial state after a training, and modify the weights in-place,

In that case, `reset` is not strictly necessary. We can just decide mutable pipelines, indicating in documentation how to apply multiple trainings over same configuration:

````python
for train_id in range(1,10):
  pl = Pipeline.from_config(....)
  ...
  pl.train(...)
```` 

> or moving the inference methods to the _PreTrainedPipeline class, although this will break the possibility of calling the predict method to test a pipeline configuration.

We need to keep the "mock prediction" for now, at least until design a better path for data annotation sessions.


You can go ahead with this approach (mutable pipeline), but keeping predictions fro non trained pipelines.

dvsrepo:
> > In summary i vote for either:
> > implementing some sort of reset method that allows to reset a model to its initial state after a training, and modify the weights in-place,
> 
> In that case, `reset` is not strictly necessary. We can just decide mutable pipelines, indicating in documentation how to apply multiple trainings over same configuration:
> 
> ```python
> for train_id in range(1,10):
>   pl = Pipeline.from_config(....)
>   ...
>   pl.train(...)
> ```
> 
> > or moving the inference methods to the _PreTrainedPipeline class, although this will break the possibility of calling the predict method to test a pipeline configuration.
> 
> We need to keep the "mock prediction" for now, at least until design a better path for data annotation sessions.
> 
> You can go ahead with this approach (mutable pipeline), but keeping predictions fro non trained pipelines.

👍

dcfidalgo:
Closing this one, see #462

PR #411, opened 2020-10-19
Avoid copy of pipeline when running a learning rate search

dcfidalgo:
This PR improves a bit the `Pipeline.find_lr` method. Instead of copying the whole pipeline, it saves the model state before the learning rate scan and restores it afterwards.

### DIFF
--- a/src/biome/text/_helpers.py
+++ b/src/biome/text/_helpers.py
@@ -286,7 +286,7 @@ def create_dataloader(
 
 
 def create_trainer_for_finding_lr(
-    pipeline: Pipeline,
+    model: PipelineModel,
     trainer_config: TrainerConfiguration,
     training_data: InstancesDataset,
 ) -> GradientDescentTrainer:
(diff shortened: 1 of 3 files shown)


frascuchon (approved):
Nice !!!

Could we use this approach in pipeline train method ??

I imagine some kind of backup mechanism in same way, like:
```python
with model_backup(): ## this with statement will restore pipeline after the train
   # do the train
  
```

dcfidalgo:
Ah this is actually a very nice idea! In #406 i advocate that we modify the weights in-place to have a more straight forward inference flow. But we could provide the `Pipeline.train` method also a context manger so that `Pipeline.train()` would modify the parameters in-place, and `with Pipeline.train()` would restore the model state afterwards. @frascuchon what do you think?

dcfidalgo:
Sorry, my previous comment was actually bull****!
I would imagine something like this:
pl = Pipeline.from_...
with model_backup():
    pl.train()
```
@frascuchon Is this inline with your idea?

Issue #403, opened 2020-10-16
Implement cache mechanism for the instances in our `Dataset` class


dcfidalgo:
TODO: We want to take advantage of the caching done in `datasets.Dataset` to reliably cache the created instances for a given pipeline (featurizer). For large datasets, creating the instances can take quite some time. The fingerprint of the cached dataset should include the state of the dataset as well as the "config" of the featurizer, something like that.

PR #413, opened 2020-10-21
Adding a new HPO component, making is compatible with Datasets

dcfidalgo:
This PR introduces a new HPO class (`RayTuneTrainable`) that is compatible with Datasets and is intended to replace the `HpoParams` and `HpoExperiment` classes.

Personally i find it confusing that you can define the same parameters in the `tune.Experiment` and in `tune.run`, but the latter will be ignored if they were specified in the former. So the new implementation does not make use of the `tune.Experiment` and relies more on the parameters provided directly to `tune.run`. Also working with @ignacioct we noticed that it is more intuitive and faster to just copy the configs of your pipeline or trainer, and replace the parameters with the search spaces, than to write a new dict with only the search spaces. So the merging capabilities of the `HpoParams` are not really needed.

A minimal usage example of the new class would be:
```python
my_trainable = RayTuneTrainable(pipeline_config, trainer_config, train_dataset, valid_dataset)
tune.run(my_trainable.func, config=my_trainable.config)
```
For a more detailed usage you can have a look at the updated tutorial.
I think it's a bit more elegant than the three step flow we have right now:
```
HpoParams -> HpoExperiment -> ray.tune(HpoExperiment.as_tune_experiment())
```

@frascuchon If you are OK with this proposal i would go ahead and remove the old HPO components in a follow-up PR.

### DIFF
--- a/docs/docs/documentation/tutorials/3-Hyperparameter_optimization_with_Ray_Tune.ipynb
+++ b/docs/docs/documentation/tutorials/3-Hyperparameter_optimization_with_Ray_Tune.ipynb
@@ -93,7 +93,7 @@
     "from biome.text import Pipeline, Dataset\n",
     "from biome.text.configuration import VocabularyConfiguration, WordFeatures, TrainerConfiguration\n",
     "from biome.text.loggers import BaseTrainLogger\n",
-    "from biome.text.hpo import RayTuneTrainable\n",
+    "from biome.text.hpo import TuneExperiment\n",
     "from ray import tune"
    ]
   },
@@ -108,7 +108,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "As a first step we will download the training and validation data to our local machine, and create our dataset. "
+    "As a first step we will download the training and validation data to our local machine, and create our datasets. "
    ]
   },
   {
@@ -135,128 +135,15 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Creating the pipeline and vocabulary"
+    "## Defining the pipeline and the search space"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "In order to be more efficient and speed things up, we will create the vocabulary beforehand and reuse it in every trial.\n",
-    "For this we have to define a `Pipeline` first and create the vocabulary from our datasets.\n",
+    "As mentioned in the introduction we will use the same pipeline configuration as used in the [base tutorial](https://www.recogn.ai/biome-text/documentation/tutorials/1-Training_a_text_classifier.html)).\n",
     "\n",
-    "Let's start with defining the configuration of our pipeline (for details see the [base tutorial](https://www.recogn.ai/biome-text/documentation/tutorials/1-Training_a_text_classifier.html)):"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "labels = list(set(train_ds[\"label\"]))"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "pipeline_config = {\n",
-    "    \"name\": \"german_business_names\",\n",
-    "    \"tokenizer\": {\n",
-    "        \"text_cleaning\": {\n",
-    "            \"rules\": [\"strip_spaces\"]\n",
-    "        }\n",
-    "    },\n",
-    "    \"features\": {\n",
-    "        \"word\": {\n",
-    "            \"embedding_dim\": 64,\n",
-    "            \"lowercase_tokens\": True,\n",
-    "        },\n",
-    "        \"char\": {\n",
-    "            \"embedding_dim\": 32,\n",
-    "            \"lowercase_characters\": True,\n",
-    "            \"encoder\": {\n",
-    "                \"type\": \"gru\",\n",
-    "                \"num_layers\": 1,\n",
-    "                \"hidden_size\": 32,\n",
-    "                \"bidirectional\": True,\n",
-    "            },\n",
-    "            \"dropout\": 0.1,\n",
-    "        },\n",
-    "    },\n",
-    "    \"head\": {\n",
-    "        \"type\": \"TextClassification\",\n",
-    "        \"labels\": labels,\n",
-    "        \"pooler\": {\n",
-    "            \"type\": \"gru\",\n",
-    "            \"num_layers\": 1,\n",
-    "            \"hidden_size\": 32,\n",
-    "            \"bidirectional\": True,\n",
-    "        },\n",
-    "        \"feedforward\": {\n",
-    "            \"num_layers\": 1,\n",
-    "            \"hidden_dims\": [32],\n",
-    "            \"activations\": [\"relu\"],\n",
-    "            \"dropout\": [0.0],\n",
-    "        },\n",
-    "    },       \n",
-    "}"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "We will use this configuration dictionary to create our pipeline:"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "pl = Pipeline.from_config(pipeline_config)"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "Next, we have to define the vocabulary configuration and create our vocabulary.\n",
-    "\n",
-    "\n",
-    "::: tip Note\n",
-    "\n",
-    "If you want to optimize the vocabulary configuration in the hyperparameter search (for example, the `min_count` argument), you have to move the vocabulary creation to the `trainable` function below. That is, in each trial the vocabulary will be created anew.\n",
-    "\n",
-    ":::"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "vocab_config = VocabularyConfiguration(sources=[train_ds], min_count={WordFeatures.namespace: 20})\n",
-    "pl.create_vocabulary(vocab_config)"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "## Defining the search space"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
     "To perform a random hyperparameter search (as well as a grid search) we simply have to replace the parameters we want to optimize with methods from the [Random Distributions API](https://docs.ray.io/en/latest/tune/api_docs/search_space.html#random-distributions-api) or the [Grid Search API](https://docs.ray.io/en/latest/tune/api_docs/search_space.html#grid-search-api) in our configuration dictionaries.\n",
     "For a complete description of both APIs and how they interplay with each other, see the corresponding section in the [Ray Tune docs](https://docs.ray.io/en/latest/tune/api_docs/search_space.html). \n",
     "\n",
@@ -315,7 +202,7 @@
     "    },\n",
     "    \"head\": {\n",
     "        \"type\": \"TextClassification\",\n",
-    "        \"labels\": labels,\n",
+    "        \"labels\": list(set(train_ds[\"label\"])),\n",
     "        \"pooler\": {\n",
     "            \"type\": tune.choice([\"gru\", \"lstm\"]),\n",
     "            \"num_layers\": tune.choice([1, 2]),\n",
@@ -364,38 +251,17 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Now we have everything ready to start our random hyperparameter search.\n",
-    "The `tune.run` method requires a trainable function that takes as argument a configuration.\n",
-    "*biome.text* provides both elements via the `RayTuneTrainable` class, which we instantiate with our configuration dictionaries, our datasets and a vocabulary to be shared among the trials for efficiency reasons, but this is optional.\n",
-    "We also provide a name that will mainly be used as project and experiment name for the integrated WandB and MLFlow logger, respectively."
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "trainable = RayTuneTrainable(\n",
-    "    pipeline_config=pipeline_config,\n",
-    "    trainer_config=trainer_config,\n",
-    "    train_dataset=train_ds,\n",
-    "    valid_dataset=valid_ds,\n",
-    "    vocab=pl.backbone.vocab,\n",
-    "    name=\"My first HPO experiment\"\n",
-    ")"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "The number of trials our search will go through depends on the `num_samples` parameter of the `tune.run` method.\n",
+    "Before starting our random hyperparameter search we first have to create a `TuneExperiment` instance with our configurations dicts and our datasets.\n",
+    "We also set a name that will mainly be used as project and experiment name for the integrated WandB and MLFlow logger, respectively.\n",
+    "\n",
+    "Furthermore, we can pass on all the parameters available for the underlying [`tune.Experiment`](https://docs.ray.io/en/master/tune/api_docs/execution.html#tune-experiment) class:\n",
+    "\n",
+    "- The number of trials our search will go through depends on the `num_samples` parameter.\n",
     "In our case, a random search, it equals the number of trials, whereas in the case of a grid search the total number of trials is `num_samples` times the grid configurations (see the [Tune docs](https://docs.ray.io/en/latest/tune/api_docs/search_space.html#overview) for illustrative examples).\n",
     "\n",
-    "The `local_dir` parameter defines the output directory of the HPO results and will also contain the training results of each trial (that is the model weights and metrics).\n",
+    "- The `local_dir` parameter defines the output directory of the HPO results and will also contain the training results of each trial (that is the model weights and metrics).\n",
     "\n",
-    "The number of parallel running trials depends on your `resources_per_trial` configuration and your local resources.\n",
+    "- The number of parallel running trials depends on your `resources_per_trial` configuration and your local resources.\n",
     "The default value is `{\"cpu\": 1, \"gpu\": 0}` and results, for example, in 8 parallel running trials on a machine with 8 CPUs.\n",
     "You can also use fractional values. To share a GPU between 2 trials, for example, pass on `{\"gpu\": 0.5}`. \n",
     "\n",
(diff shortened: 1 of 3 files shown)


ignacioct:
I think this makes the flow of the optimization a bit more fluid, and there are less potentially redundant elements on the script.

dvsrepo:
In principle, I agree this workflow is simpler, but just a few comments:

- I think that for multistep tuning workflows the config merging feature is
quite nice and cleaner than copying and passing the whole config (see
tutorial for the Automl summit on our Dev jupyterlab). In the current hpo
tutorial we have this huge config dict twice just to create the vocab and
then do the hpo with 9 params. I am not strongly against this but just to
mention we should take into account more complex workflows.

- in terms of design I think using ray.Experiment is also pretty clean and
standard. Maybe is a matter of streamlining the API (see below).

- I agree it's a bit misleading to have to specify the config when creating
the trainable and in the tune run.

- we've never tested this but using Experiment you can pass a list of
experiments to the ray tune run function. This is useful for creating
groups of potentially incompatible configs (e.g. tuning some params with
interdependencies that will make some sample configs to be incorrect) on
one run (this is specially relevant when running larger hpo against a
cluster for example, you would like to avoid doing two or three tune.run
executions for the same problem target). With this implementation we loose
this.

- Maybe this workflow simplification could be achieved by keeping
HpoExperiment (or TuneExperiment which might be a better name) with the
ability to pass the params directly removing the need of the extra
hpoparams class?


- the original idea, I think, was to have our hpo abstraction to
accommodate for other tuning libraries, thus the .as_tune step but given
the uptake of ray tune I think is safe to focus on ray tune exclusively to
simplify things (as you are trying to do here).

- not for this PR and not covered by previous impl either, but please think
about how could we do the hp tuning with pretrained pipelines (tuning the
head and the trainer for example) .

- finally, let's have a minimal version you all agree now, taking into
account we will be exploring having raysgd or lightning as trainer soon.

PS:
- for the tutorial I would avoid the vocab creation step, to make it more
straightforward to read and focus on the tuning part.

dcfidalgo:
Hey Dani, thanks a lot for your input! I will try to elaborate i bit more my thought process for the new class.

**experiment**

I guess the reason why they exposed the same parameters twice (in `tune.Experiment` and `tune.run`) is to provide a short cut for the most common use cases. So instead of having to do `trainable + config -> Experiment -> tune.run` you can go `trainable + config -> tune.run`. So why not take the short cut.
> - in terms of design I think using ray.Experiment is also pretty clean and
standard. Maybe is a matter of streamlining the API (see below).

Maybe that was my starting point. I think we are making our lives too complicated by having to streamline the `Experiment` API to `HpoExperiment` without really getting a cleaner interface, since we still have to go through `tune.run`. For example, i wanted to specify the `resources_per_trial` to use the GPU, but `HpoExperiment` does not have this parameter. However `tune.run` does. So i specified it in `tune.run`, but of course this is not how it works. Once you pass in a `tune.Experiment` to `tune.run` all the corresponding parameters get ignored.

So instead of having to streamline the `tune.Experiment` API i would just directly use the `tune.run` API.

I guess there are two use cases where you really want to use the `tune.Experiment` class and not take the short cut mentioned above:
- You want to launch several Experiments at the same time. Then i think it's actually cleaner to use `tune.run_experiments` (which does not expose all `tune.Experiment` parameters) than `tune.run`;
- You want to reuse the same experiment for several HPO runs to try out different schedulers, for example.

We have not encountered these use case yet, but a `TuneExperiment` class would indeed be beneficial here. With the new approach, however, it is also not too complicated: 
```python
my_trainable = RayTuneTrainable(...)
my_experiment = tune.Experiment(my_trainable.func, my_trainable.config)
```

**vocab**

Yes, i totally agree that having to create a Pipeline just to share the vocab is an overkill, i will remove this part from the tutorial. Actually i had a TODO on the `vocab` argument in the new `RayTuneTrainable` class saying that this option will likely go away. The reason is that @frascuchon and I are trying to find a working caching mechanism for the instances #403, and right now building the vocab is slow because we also have to build the instances. Once we can load the cached instances, there is no real gain in reusing the vocab among the trials.

**config merging**

Here i did not know how to elegantly implement this. Requiring a `Pipeline` just to use its config seemed an overkill to me, and having to specify two configs (something like `pipeline_config` + `pipeline_search_space_config`) seemed confusing ... maybe `pipeline_config` could be marked as optional, not sure. Also having this merging capabilities only for the `pipeline_config` but not for the `trainer_config` seemed inconsistent. So i thought that if the user wants to merge or update configs he/she could do this outside of the `RayTuneTrainable` class:
```python
initial_pl_config = deepcopy(pl_config)
pl_config.update({"features": {...})
# or
pl_config["features"]["char"] = {...}

my_trainable = RayTuneTrainable(pipeline_config=pl_config, ...)
```
--------------------------
> - not for this PR and not covered by previous impl either, but please think
about how could we do the hp tuning with pretrained pipelines (tuning the
head and the trainer for example) .

I think this is already possible when using transformers for example, where you can set `trainable = False`. @frascuchon Maybe we could add such an option to all the features and the encoder?
----------------
woa, way too long this comment ... will shut up now.

dvsrepo:
> Hey Dani, thanks a lot for your input! I will try to elaborate i bit more my thought process for the new class.
> 
> **experiment**
> 
> I guess the reason why they exposed the same parameters twice (in `tune.Experiment` and `tune.run`) is to provide a short cut for the most common use cases. So instead of having to do `trainable + config -> Experiment -> tune.run` you can go `trainable + config -> tune.run`. So why not take the short cut.
> 
> > * in terms of design I think using ray.Experiment is also pretty clean and
> >   standard. Maybe is a matter of streamlining the API (see below).
> 
> Maybe that was my starting point. I think we are making our lives too complicated by having to streamline the `Experiment` API to `HpoExperiment` without really getting a cleaner interface, since we still have to go through `tune.run`. For example, i wanted to specify the `resources_per_trial` to use the GPU, but `HpoExperiment` does not have this parameter. However `tune.run` does. So i specified it in `tune.run`, but of course this is not how it works. Once you pass in a `tune.Experiment` to `tune.run` all the corresponding parameters get ignored.
> 
> So instead of having to streamline the `tune.Experiment` API i would just directly use the `tune.run` API.
> 
> I guess there are two use cases where you really want to use the `tune.Experiment` class and not take the short cut mentioned above:
> 
> * You want to launch several Experiments at the same time. Then i think it's actually cleaner to use `tune.run_experiments` (which does not expose all `tune.Experiment` parameters) than `tune.run`;
> * You want to reuse the same experiment for several HPO runs to try out different schedulers, for example.
> 
> We have not encountered these use case yet, but a `TuneExperiment` class would indeed be beneficial here. With the new approach, however, it is also not too complicated:
> 
> ```python
> my_trainable = RayTuneTrainable(...)
> my_experiment = tune.Experiment(my_trainable.func, my_trainable.config)
> ```
> 
Thanks for the explanation, given this I agree, we do not have to enforce the Experiment API.

> **vocab**
> 
> Yes, i totally agree that having to create a Pipeline just to share the vocab is an overkill, i will remove this part from the tutorial. Actually i had a TODO on the `vocab` argument in the new `RayTuneTrainable` class saying that this option will likely go away. The reason is that @frascuchon and I are trying to find a working caching mechanism for the instances #403, and right now building the vocab is slow because we also have to build the instances. Once we can load the cached instances, there is no real gain in reusing the vocab among the trials.

Yes, this is something we need to aim for.

> 
> **config merging**
> 
> Here i did not know how to elegantly implement this. Requiring a `Pipeline` just to use its config seemed an overkill to me, and having to specify two configs (something like `pipeline_config` + `pipeline_search_space_config`) seemed confusing ... maybe `pipeline_config` could be marked as optional, not sure. Also having this merging capabilities only for the `pipeline_config` but not for the `trainer_config` seemed inconsistent. So i thought that if the user wants to merge or update configs he/she could do this outside of the `RayTuneTrainable` class:
> 
> ```python
> initial_pl_config = deepcopy(pl_config)
> pl_config.update({"features": {...})
> # or
> pl_config["features"]["char"] = {...}
> 
> my_trainable = RayTuneTrainable(pipeline_config=pl_config, ...)
> ```
> 
OK
> > * not for this PR and not covered by previous impl either, but please think
> >   about how could we do the hp tuning with pretrained pipelines (tuning the
> >   head and the trainer for example) .
> 
> I think this is already possible when using transformers for example, where you can set `trainable = False`. @frascuchon Maybe we could add such an option to all the features and the encoder?
> 
Yes, with the `transformers` feature is kind of possible, even with `trainable = True` because you have `model_name` pointing to the pre-trained model but I was referring to features+encoder as loaded with `Pipeline.from_pretrained()`. In my previous comment, I was referring to the head but this is a more general thing. Imagine a pretrained model you want to fine-tune for the same task but with new data and you want to do HPO for the trainer. If its a transfomer-based model you would save the model to disk and then use that model ref in the features section, but even in that case you will be training the head from scratch. What I am trying to say is that this "tuning process" is no a one-shot thing. As you can see, this is related to passing the pipeline to the Tuner.

> woa, way too long this comment ... will shut up now.

dcfidalgo:
>  In my previous comment, I was referring to the head but this is a more general thing. Imagine a pretrained model you want to fine-tune for the same task but with new data and you want to do HPO for the trainer. If its a transfomer-based model you would save the model to disk and then use that model ref in the features section, but even in that case you will be training the head from scratch. What I am trying to say is that this "tuning process" is no a one-shot thing. As you can see, this is related to passing the pipeline to the Tuner.

Ah, yes, now i get it. For this maybe `pipeline_config` could be either a dict or a string pointing to a `model.tar.gz` and the trainable function would use `Pipeline.from_pretrained` in that case.

frascuchon (commented):
For me it's ok the change once the workflow is more clear.

My main comment is about `Experiment` class. I prefer to keep using it for several reasons:

* `tune.run` works internally with `Experiment` objects, creating them if you pass raw arguments.

* `Experiment` class is the main data model that ray uses for trials configurations, so keeping close to it, will simplify futures integrations like RaySVG or more complex workflow executions.

* `tune.run` is, effectively, a shortcut for launch trails easily, but the appropriate way should be use the `ray.run_experiments` method.

* `Experiment` class expose resource arguments that `tune.run` does, so we should have no problem controlling that type of configuration.

We could make the `RayTuneTrainable` extends from `Experiment` and use directly in tune methods:

```python
trainable = RayTuneTrainable(....)
tune.run (trainable,....)

# or
tune.run_experiments(trainable,...)
```

Having this abstraction layer will simplify futures modifications or features over hpo workflow, and is strongly recommended do it like this.


In terms of vocabulary creation, we will have improvement with dataset caching, but since ray is a distributed environment, this improvement won't match for all cases. Even in local environments with a huge dataset, an initial vocab creation will be a better options (the vocabulary creation over allennlp instances computing time can be a serious bottleneck.

dcfidalgo:
Ok, how about we keep the `to_experiment` or `as_experiment` method in the `RayTuneTrainable` and accept as argument a `**expriment_kwargs` that is passed on to `tune.Experiment`. In this way you can choose which one of the flows you prefer:
```python
tune.run(my_trainable.func, config=my_trainable.config, **experiment_kwargs)
# or
tune.run_experiment(my_trainable.to_experiment(**experiment_kwargs))
```
Or move the `experiment_kwargs` to the `RayTuneTrainable.__init__` to be able to simply call 
```python
tune.run_experiment(my_trainable.to_experiment())
```

> In terms of vocabulary creation, we will have improvement with dataset caching, but since ray is a distributed environment, this improvement won't match for all cases. Even in local environments with a huge dataset, an initial vocab creation will be a better options (the vocabulary creation over allennlp instances computing time can be a serious bottleneck.

Yes, i do not really have numbers here, i will try to benchmark a bit the "vocabulary with instances" creation to estimate the gain.

dcfidalgo:
Just did some massive featurizing and vocab creating ...
some numbers:

**Arxiv dataset**

- 1e4 examples with an average of 977 chars, for our standards an average dataset i would say.
- Instance creation, 1 cpu: 1 min 22 s +- 2s
- vocab creation, 1 cpu: 3.8 s +- 0.5 s

**Wikipedia**

- 1e5 examples with an average of 2500 chars, far from massive but for our standards impressive i would say
- Instance creation using the 12 cpus on my 6 core machine: ~15 min
- vocab creation, 1 cpu: 1 min 53 s +- 9 s

Given the fact that a trial for such big datasets would take at least on the scale of hours, not sure if the vocab creation would really be an issue. If you have sufficient CPUs to create the instances, i would even say that the instance creation is actually not really a bottleneck.

-------

On the other hand i followed @frascuchon suggestion on making the `RayTuneTrainable` a child class of `tune.Experiment`. This actually looks really clean with following init:
```python
    def __init__(
        self,
        pipeline_config: dict,
        trainer_config: dict,
        train_dataset: Dataset,
        valid_dataset: Dataset,
        vocab: Optional[Vocabulary] = None,
        name: Optional[str] = None,
        **experiment_kwargs,
    ):
```
where experiment_kwargs is passed on to `tune.Experiment`. I tried and `tune.run(my_experiment, ...)` and `tune.run_experiments(my_experiment, ...)` works without issues. `tune.run_experiments`, however, does not return an Analysis object, but a list of Trial objects which is far from optimal to figure out the best performing run ... when you look at the issues in github it sounds like they want to get rid of `tune.run_experiments` (https://github.com/ray-project/ray/issues/8127, https://github.com/ray-project/ray/pull/4515).

So following your comments my proposal would be to:
- rename `RayTuneTrainable` -> `TuneExperiment`
- inherit from `tune.Experiment`
- rename `TuneExperiment.func` to `TuneExperiment.trainable`

This allows for two flows:
```python
my_experiment = TuneExperiment(..., resources_per_trial={...})
analysis = tune.run(my_experiment)
# or
my_experiment = TuneExperiment(..., resources_per_trial={...})
analysis = tune.run(my_experiment.trainable, config=my_experiment.config, resources_per_trial={...})
```
The second flow could be useful if you want to use `TuneExperiment.config` but with a custom trainable function. Or we add another parameter in `TuneExperiment` that takes a callable, like we do it now in our `HpoExperiment` with `trainable_fn`. Actually i think i prefer latter.

@frascuchon would you be ok with these changes?

dvsrepo:
> Just did some massive featurizing and vocab creating ...
> some numbers:
> 
> **Arxiv dataset**
> 
> * 1e4 examples with an average of 977 chars, for our standards an average dataset i would say.
> * Instance creation, 1 cpu: 1 min 22 s +- 2s
> * vocab creation, 1 cpu: 3.8 s +- 0.5 s
> 
> **Wikipedia**
> 
> * 1e5 examples with an average of 2500 chars, far from massive but for our standards impressive i would say
> * Instance creation using the 12 cpus on my 6 core machine: ~15 min
> * vocab creation, 1 cpu: 1 min 53 s +- 9 s
> 
> Given the fact that a trial for such big datasets would take at least on the scale of hours, not sure if the vocab creation would really be an issue. If you have sufficient CPUs to create the instances, i would even say that the instance creation is actually not really a bottleneck.
> 
> On the other hand i followed @frascuchon suggestion on making the `RayTuneTrainable` a child class of `tune.Experiment`. This actually looks really clean with following init:
> 
> ```python
>     def __init__(
>         self,
>         pipeline_config: dict,
>         trainer_config: dict,
>         train_dataset: Dataset,
>         valid_dataset: Dataset,
>         vocab: Optional[Vocabulary] = None,
>         name: Optional[str] = None,
>         **experiment_kwargs,
>     ):
> ```
> 
> where experiment_kwargs is passed on to `tune.Experiment`. I tried and `tune.run(my_experiment, ...)` and `tune.run_experiments(my_experiment, ...)` works without issues. `tune.run_experiments`, however, does not return an Analysis object, but a list of Trial objects which is far from optimal to figure out the best performing run ... when you look at the issues in github it sounds like they want to get rid of `tune.run_experiments` ([ray-project/ray#8127](https://github.com/ray-project/ray/issues/8127), [ray-project/ray#4515](https://github.com/ray-project/ray/pull/4515)).
> 
> So following your comments my proposal would be to:
> 
> * rename `RayTuneTrainable` -> `TuneExperiment`
> * inherit from `tune.Experiment`
> * rename `TuneExperiment.func` to `TuneExperiment.trainable`
> 
> This allows for two flows:
> 
> ```python
> my_experiment = TuneExperiment(..., resources_per_trial={...})
> analysis = tune.run(my_experiment)
> # or
> my_experiment = TuneExperiment(...)
> analysis = tune.run(my_experiment.trainable, config=my_experiment.config, resources_per_trial={...})
> ```
> 
> The second flow could be useful if you want to use `TuneExperiment.config` but with a custom trainable function. Or we add another parameter in `TuneExperiment` that takes a callable, like we do it now in our `HpoExperiment` with `trainable_fn`. Actually i think i prefer latter.

💯 this looks good to me!

> @frascuchon would you be ok with these changes?

dcfidalgo:
I pushed the suggestions discussed above.

Today i was fighting the whole day with a weird issue. I did not manage to run the tutorial with ipython>=7.0.0 and the corresponding jupyter-console and prompt-toolkit. Only with ipython==6.5.0 did i succeed.
This is the traceback:
```python
(pid=807256) Traceback (most recent call last):
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/ray/function_manager.py", line 493, in _load_actor_class_from_gcs
(pid=807256)     actor_class = pickle.loads(pickled_class)
(pid=807256)   File "/home/david/recognai/biome/biome-text/src/biome/text/__init__.py", line 33, in <module>
(pid=807256)     from .pipeline import Pipeline
(pid=807256)   File "/home/david/recognai/biome/biome-text/src/biome/text/pipeline.py", line 24, in <module>
(pid=807256)     from biome.text.configuration import (
(pid=807256)   File "/home/david/recognai/biome/biome-text/src/biome/text/configuration.py", line 19, in <module>
(pid=807256)     from .modules.heads.task_head import TaskHeadConfiguration
(pid=807256)   File "/home/david/recognai/biome/biome-text/src/biome/text/modules/heads/__init__.py", line 3, in <module>
(pid=807256)     from .classification.doc_classification import (
(pid=807256)   File "/home/david/recognai/biome/biome-text/src/biome/text/modules/heads/classification/doc_classification.py", line 10, in <module>
(pid=807256)     from captum.attr import IntegratedGradients
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/captum/attr/__init__.py", line 42, in <module>
(pid=807256)     from ._utils import visualization  # noqa
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/captum/attr/_utils/visualization.py", line 7, in <module>
(pid=807256)     from matplotlib import pyplot as plt
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/matplotlib/pyplot.py", line 2336, in <module>
(pid=807256)     switch_backend(rcParams["backend"])
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/matplotlib/pyplot.py", line 276, in switch_backend
(pid=807256)     class backend_mod(matplotlib.backend_bases._Backend):
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/matplotlib/pyplot.py", line 277, in backend_mod
(pid=807256)     locals().update(vars(importlib.import_module(backend_name)))
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/importlib/__init__.py", line 127, in import_module
(pid=807256)     return _bootstrap._gcd_import(name[level:], package, level)
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/ipykernel/__init__.py", line 2, in <module>
(pid=807256)     from .connect import *
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/ipykernel/connect.py", line 13, in <module>
(pid=807256)     from IPython.core.profiledir import ProfileDir
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/IPython/__init__.py", line 56, in <module>
(pid=807256)     from .terminal.embed import embed
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/IPython/terminal/embed.py", line 16, in <module>
(pid=807256)     from IPython.terminal.interactiveshell import TerminalInteractiveShell
(pid=807256)   File "/home/david/miniconda3/envs/biome/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 21, in <module>
(pid=807256)     from prompt_toolkit.formatted_text import PygmentsTokens
(pid=807256) ModuleNotFoundError: No module named 'prompt_toolkit.formatted_text'
```
Thought i would write it down here to keep it somewhere.

frascuchon (approved):

PR #421, opened 2020-10-27
refactor: using exclusive or outputs features for Token classifier

frascuchon:
This PR splits  `labels` field in `TokenClassification` head in `labels` (tag labels)  and `entities` (span labels). Consequently, multiple output can be defined in pipelines, returning a list of names, instead of a single string.

### DIFF
--- a/src/biome/text/dataset.py
+++ b/src/biome/text/dataset.py
@@ -276,7 +276,9 @@ class Dataset:
         """
         self._LOGGER.info("Creating instances ...")
 
-        input_columns = [col for col in pipeline.inputs + pipeline.output if col]
+        input_columns = [(col, False) for col in pipeline.inputs if col]
+        input_columns.extend([(col, True) for col in pipeline.output if col])
+
         dataset_with_pickled_instances = self.dataset.map(
             self._create_and_pickle_instances,
             fn_kwargs={
@@ -287,7 +289,8 @@ class Dataset:
             # trying to be smart about multiprocessing,
             # at least 1000 examples per process to avoid overhead,
             # but 1000 is a pretty random number, can surely be optimized
-            num_proc=num_proc or min(
+            num_proc=num_proc
+            or min(
                 max(1, int(len(self.dataset) / 1000)),
                 int(multiprocessing.cpu_count() / 2),
             ),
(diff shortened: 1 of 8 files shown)


tests/text/modules/heads/test_token_classification.py
```diff
@@ -60,6 +60,8 @@ def trainer_dict() -> Dict:
 def test_train(pipeline_dict, training_data_source, trainer_dict, tmp_path):
     pipeline = Pipeline.from_config(pipeline_dict)
 
+    assert pipeline.output == ["entities", "labels"]
```
dcfidalgo, line 63: This is a bit confusing. Right now the `pipeline.output` are the optional arguments of the `featurize` method of the head. But the `pipeline.predict` does not return a `labels`, but a `tags` ... maybe we should somehow enforce a consistency between the two things, or better define what `pipeline.output` actually means.

dcfidalgo (commented):
Nice, much cleaner this way! 
I have a slight preference for 'tags' instead of 'labels' in the `featurize` method, but no strong opinion.
I just left a couple of inline comments.

tests/text/modules/heads/test_token_classification.py
frascuchon: For now, I've updated the `featurize` method, renaming `labels` to `tags`.   

We must define a data model for task input/outputs that will keep this logic, reducing this kind of frictions/confussions.

src/biome/text/dataset.py
```diff
@@ -312,7 +312,7 @@ def to_instances(self, pipeline: "Pipeline", lazy=True, num_proc: Optional[int]
     @staticmethod
     def _create_and_pickle_instances(row, input_columns, featurize, instances_col_name):
         """Helper function to be used together with the `datasets.Dataset.map` method"""
-        instance = featurize(**{key: row[key] for key in input_columns})
+        instance = featurize(**{key: row.get(key) for key in input_columns})
```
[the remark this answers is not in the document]
frascuchon: Again, include data model for input and outputs we can easily handle this logic. 

Anyway, I've made some changes for take into account optional/non optionals inputs.

dcfidalgo (approved):
Nice, LGTM!

PR #422, opened 2020-10-28
fix(token-class): aligns raw text and tags tokenization

frascuchon:
This PR fixes misaligned tokenization for `TokenClassifier` between `raw_text` and `tags`, when original text contains spaces symbols like `\n`.

### DIFF
--- a/src/biome/text/helpers.py
+++ b/src/biome/text/helpers.py
@@ -313,9 +313,6 @@ def tags_from_offsets(
     -------
     tags (BIOUL or BIO)
     """
-    if offsets is None:
-        return []
-
     tags = spacy.gold.biluo_tags_from_offsets(
         doc, [(offset["start"], offset["end"], offset["label"]) for offset in offsets]
     )
--- a/src/biome/text/helpers.py
+++ b/src/biome/text/helpers.py
@@ -7,6 +7,7 @@ from inspect import Parameter
 from typing import Any, Callable, Dict, Iterable, List, Tuple, Type, Optional
 
 import spacy
+import spacy.gold
 import yaml
 from allennlp.common import util
 from elasticsearch import Elasticsearch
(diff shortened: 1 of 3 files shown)


src/biome/text/helpers.py
```diff
@@ -312,6 +313,9 @@ def tags_from_offsets(
     -------
     tags (BIOUL or BIO)
     """
+    if offsets is None:
+        return []
+
```
dcfidalgo, line 318: I would move this check to the caller of this function, see comment below
```suggestion
```

src/biome/text/modules/heads/token_classification.py
```diff
@@ -142,28 +140,31 @@ def featurize(
 
             They are used with the `spacy.gold.biluo_tags_from_offsets` method.
         """
+        if isinstance(text, str):
+            doc = self.backbone.tokenizer.nlp(text)
+            tokens = [token.text for token in doc]
+            tags = tags_from_offsets(doc, labels, self._label_encoding)
```
dcfidalgo, line 146: Moved check here:
```suggestion
            tags = tags_from_offsets(doc, labels, self._label_encoding) if labels is not None else []
```

src/biome/text/modules/heads/token_classification.py
```diff
@@ -142,28 +140,31 @@ def featurize(
 
             They are used with the `spacy.gold.biluo_tags_from_offsets` method.
         """
+        if isinstance(text, str):
+            doc = self.backbone.tokenizer.nlp(text)
+            tokens = [token.text for token in doc]
```
dcfidalgo, line 145: This basically substitutes the tokenization step otherwise done in `self.backbone.featurizer(text, to_field="text", tokenize=True, aggregate=True)`, right? However, in `self.backbone.featurizer` we do a lot of checks on how to exactly tokenize the data, not sure how this influences all the different possibilities ... did you check this?

dcfidalgo (commented):
LGTM just a few inline comments

src/biome/text/modules/heads/token_classification.py
frascuchon: For token classification, if we must tokenize the original text, we must control the whole tokenization process, keeping the original text accesible once the tokenization is done.

Our tokenizer incude pre and post tasks for text and token cleaning that we must by-pass in token classification. We continue using tokenizer featurization, just skipping tokenization phase, like with pre-tokenized text.

PR #423, opened 2020-10-28
fix: default explain not working for token classification

frascuchon:
This PR allow execute pipeline explain for token classification, processing `SequenceLabelField` and `MetadataField` fields types


src/biome/text/modules/heads/task_head.py
```diff
@@ -141,7 +141,7 @@ def explain_prediction(
         -------
             Prediction with explanation
         """
-        raise {**prediction, "explain": {}}
+        return {**prediction, "explain": {}}
```
dcfidalgo, line 144: hihi

dcfidalgo (approved):
LGTM

PR #427, opened 2020-10-30
Fix/tests datasets

ignacioct:
test_features_transformers, test_text_classification_test, test_pipeline_with_custom_head and test_find_lr updated to work alongside Huggingface Datasets.

Also, gitignore updated to exclude .pylintrc, pylint configuration files (usually autoproduced when user changes configuration, could be useful to exclude it).


.gitignore
```diff
@@ -136,3 +136,6 @@ generated*
 **/mlruns/
 **/runs/
 **/.yalc/
+
+#pylint
+.pylintrc
```
dcfidalgo, line 141: I have no idea what this file does ... @frascuchon do we still need this? Can we safely delete it?
Anyways, i don't think we should mention it in `.gitignore`:
```suggestion
```

tests/integration/test_text_classification.py
```diff
@@ -7,25 +7,27 @@
...
-from biome.text import Pipeline, TrainerConfiguration, VocabularyConfiguration
+from biome.text import Pipeline, Dataset, TrainerConfiguration, VocabularyConfiguration
 from biome.text.configuration import CharFeatures
 from biome.text.configuration import WordFeatures
-from biome.text.data import DataSource
 
 
 @pytest.fixture
-def train_valid_data_source() -> Tuple[DataSource, DataSource]:
+def train_valid_data_source() -> Tuple[Dataset, Dataset]:
```
dcfidalgo, line 16: For consistency reason i would rename this fixture to `train_valid_dataset`
```suggestion
def train_valid_dataset() -> Tuple[Dataset, Dataset]:
```

tests/text/test_features_transformers.py
```diff
@@ -43,74 +48,83 @@ def pipeline_dict() -> dict:
...
+    # Test vocabulary from a pretrained file
     pl = Pipeline.from_pretrained(str(output / "model.tar.gz"))
 
+    # Check a fixed vocabulary size for the model after loading
     assert pl.backbone.vocab.get_vocab_size("transformers") == 50265
 
 
-def test_transformers_and_word(tmp_path, pipeline_dict, trainer_dict, train_data_source):
+def test_transformers_and_word(tmp_path, pipeline_dict, trainer_dict, train_dataset):
+    """ """
```
dcfidalgo, line 82: Some doc string is missing here.

tests/text/test_pipeline_with_custom_head.py
```diff
@@ -1,23 +1,37 @@
...
 
 class MyCustomHead(TextClassification):
     """Just a head renaming the original TextClassification head"""
-
     pass
 
+@pytest.fixture
+def training_dataset() -> Dataset:
+    """Creates the training dataset and gives the structure"""
+    resources_path = Path(__file__).parent.parent.parent / "tests" / "resources" / "data"
```
dcfidalgo, line 18: We could leave a TODO here:
```suggestion
    # TODO: We must cleanup the `resources` and should use a fixture to point to them
    resources_path = Path(__file__).parent.parent.parent / "tests" / "resources" / "data"
```

tests/text/test_pipeline_with_custom_head.py
```diff
@@ -1,23 +1,37 @@
...
+    training_ds = Dataset.from_csv(paths=str(resources_path / "dataset_source.csv"))
+
+    # Keeping just 'label' and text 'category'
+    training_ds = training_ds.map(
+        lambda x: {"label": x["job"], "text": x["education"] + " " + x["marital"]}, 
+        remove_columns=["education", "marital"]
+    )
+    training_ds.remove_columns_(list(set(training_ds.column_names) - set(["label", "text"])))
```
dcfidalgo, line 26: I would either remove all the columns in the `map` method, or remove the `remove_columns` argument from it. No need to remove columns twice. Actually, there is no need to remove the columns in the first place, you could just leave them in the dataset.

tests/text/test_pipeline_with_custom_head.py
```diff
@@ -34,17 +48,18 @@ def test_load_pipeline_with_custom_head():
...
-    pipeline.create_vocabulary(VocabularyConfiguration(sources=[train]))
-    pipeline.train(output=output, training=train)
+    pipeline.create_vocabulary(VocabularyConfiguration(sources=[training_dataset]))
+    pipeline.train(output=output, training=training_dataset)
 
+    # Loading model from output
     trained_pl = Pipeline.from_pretrained(os.path.join(output, "model.tar.gz"))
     trained_pl.predict("Oh yeah")
+
+    #Asserting that the pipeline head is still instance after saving and loading
```
dcfidalgo, line 64:
```suggestion
    #Asserting that the pipeline head is recognized as `MyCustomHead` instance after loading from a model.tar.gz
```

dcfidalgo (changes requested):
Nice work! Thanks a lot!

I left a few minor inline comments. @frascuchon Could you check if we still need this `.pylintrc` file?

And sorry, i did not mention that we are using [black](https://github.com/psf/black) as our code formatter, would you mind running it over the files you modified?

.gitignore
ignacioct: This is the configuration file that the Python linter pylint produces sometimes. It is a very common linter, so I thought it could be useful to add it to the .gitignore. Anyway, we can erase this change.

tests/text/test_pipeline_with_custom_head.py
ignacioct: Yeah, I'll add it. We must standarize this procedure.

tests/text/test_pipeline_with_custom_head.py
ignacioct: I like it more your way

tests/integration/test_text_classification.py
ignacioct: yes, i totally forgot to do that myself

.gitignore
dcfidalgo: ok, i see, let's wait for @frascuchon input just in case ...

frascuchon: Since we set linter configuration in `setup.cfg` I think we could drop references to `.pylintrc` file. But I have no idea what better linter config way.

ignacioct: Then, I think we can exclude this change and maybe address it later

tests/integration/test_text_classification.py
```diff
@@ -81,31 +87,35 @@ def pipeline_dict() -> dict:
...
     return {
         "batch_size": 64,
         "num_epochs": 5,
         "optimizer": {"type": "adam", "lr": 0.01},
+        "cuda_device": -1,
     }
 
 
 def test_text_classification(
     tmp_path, pipeline_dict, trainer_dict, train_valid_data_source
```
dcfidalgo, line 106: @ignacioct The CI of this PR fails because of this line, we renamed the fixture to `train_valid_dataset`. Could you push a small fix for this?

tests/integration/test_text_classification.py
```diff
@@ -81,31 +87,35 @@ def pipeline_dict() -> dict:
...
     np.random.seed(422)
     torch.manual_seed(4222)
     if torch.cuda.is_available():
         torch.cuda.manual_seed_all(4222)
 
     pl = Pipeline.from_config(pipeline_dict)
-    train_ds = pl.create_dataset(train_valid_data_source[0])
-    valid_ds = pl.create_dataset(train_valid_data_source[1])
+    train_ds = train_valid_data_source[0].to_instances(pipeline=pl)
+    valid_ds = train_valid_data_source[1].to_instances(pipeline=pl)
```
dcfidalgo, line 118: The same here

tests/integration/test_text_classification.py
ignacioct: it should be working now :)

dcfidalgo:
Ok, i will merge this now. If at some point we decide to use the `.pylintrc` file we add it again.
Thanks @ignacioct good work!

dcfidalgo (approved):

PR #428, opened 2020-10-30
Introduce caching mechanism for instance datasets

### LINKED ISSUE #403: Implement cache mechanism for the instances in our `Dataset` class

dcfidalgo:
This PR adds the capability to reuse cached instances and closes #403 . This will make the typical workflow of creating vocab and then training faster, and makes the `vocab` parameter in the `TuneExperiment` a bit superfluous as discussed in #413 .

It basically creates an identifier for an instance dataset based on:
- the fingerprint of the underlying dataset
- the tokenizer config of the pipeline
- the indexer config of the features of the pipeline
- the `biome.text.__version__`, `allennlp.__version__` and `spacy.__version__`

The last point could be optimized, since obviously most of the changes we make during development in biome.text (each commit changes `biome.text.about.__version__`) do not affect the creation of the instances. One way could be using the `__code__` attribute of the methods responsible for the instance creation. But this could become quite messy to maintain or to cover the instance creation for **all** task heads ... Also keep in mind that this is almost exclusively an issue with the development version of biome.text.

Another change worth mentioning is the introduction of an abstract `Features` class. Basically just to define a common type for all our feature classes.


src/biome/text/about.py
```diff
@@ -95,9 +95,8 @@ def version_matches(release_branch_version: str, configured_version: str) -> boo
     if release_version and not repo_tag:
         assert version_matches(release_version, version)
         return version
-    today = datetime.today().strftime("%Y%m%d%H%M%S")
 
-    return repo_tag if repo_tag else f"{version}.dev{today}+{commit}"
+    return repo_tag if repo_tag else f"{version}.dev+{commit}"
```
frascuchon, line 99: Warning: This could bring problems with package updates for dev versions, since pypi repository sort by version name (There is no sorting on commit hash)

src/biome/text/dataset.py
```diff
@@ -7,10 +7,15 @@
 from typing import Union, Dict, Iterable, List, Any, Tuple, Optional, TYPE_CHECKING
 
 import datasets
+from datasets.fingerprint import Hasher
 from allennlp.data import AllennlpDataset, AllennlpLazyDataset, Instance
+from biome.text.about import __version__ as biome__version__
```
frascuchon, line 12: `__version__` is already enabled from `biome.text` package

```suggestion
from biome.text import __version__ as biome__version__
```

src/biome/text/dataset.py
```diff
@@ -341,6 +347,33 @@ def instance_unpickler(instances_col_name: str) -> Iterable[Instance]:
...
+
+        The fingerprint is based on:
+        - the fingerprint of the previous dataset
+        - the tokenizer config
+        - the indexer config of the features
+        - the biome__version__, allennlp__version__ and spaCy__version__ just to be completely sure!
+        """
+        hasher = Hasher()
+        # add the previous fingerprint to the hash
+        hasher.update(self.dataset._fingerprint)  # necessary evil ...
```
frascuchon, line 361: The underlying fingerprint isn't built at map operation level, but at whole dataset level?

frascuchon (approved):
Nice !!! just a few questions

src/biome/text/about.py
dcfidalgo: Was not aware of that. Instead of `datetime.today` i added the **commit date** in the same format as before to the version str, this should solve the problem.

src/biome/text/dataset.py
dcfidalgo: Just figured out that these are not the same ... `from biome.text import __version__` is the biome.text version of the last install (basically when `make dev` was executed when developing) and `from biome.text.about import __version__` is the biome.text version of the last commit. I think i prefer latter one, which seems more convenient for development.

src/biome/text/dataset.py
dcfidalgo: Yes, so basically every dataset instance has a fingerprint. when applying a map operation, a new fingerprint is calculated based on the dataset's fingerprint and the arguments of the map operation. If a dataset with this new fingerprint is present in the cache, this dataset will be returned and the map operation is omitted.

src/biome/text/dataset.py
dcfidalgo: Naaa, this does not work!! `from biome.text.about import __version__` depends on the current working directory!! We have to use `from biome.text import __version__` and keep this in mind when developing.

frascuchon (approved):

PR #431, opened 2020-11-05
Fix/test pipeline explore

ignacioct:
Pipeline explore test adapted to Datasets, and explore.py changed to overcome one limitation of the explore method when declaring the dataset via dictionary


tests/text/test_pipeline_explore.py
```diff
@@ -1,38 +1,36 @@
...
 from biome.text import Pipeline
+from elasticsearch import Elasticsearch
 
 
 @pytest.fixture
-def data_source_without_mapping(tmp_path) -> DataSource:
-    data_file = tmp_path / "train.json"
-    df = pd.DataFrame({"text": ["This is a simple test"], "label": ["a"]})
-    df.to_json(data_file, lines=True, orient="records")
+def dataset_without_mapping() -> Dataset:
```
dcfidalgo, line 9: Since `Dataset`s do not have mappings, simply call this fixture `dataset`
```suggestion
def dataset() -> Dataset:
```

tests/text/test_pipeline_explore.py
```diff
@@ -1,38 +1,36 @@
...
+    """Creating the dataset"""
+    data = {"text": ["This is a simple test"], "label": ["a"]}
+    return Dataset.from_dict(data)
 
-    return DataSource(
-        source=str(data_file), flatten=False, lines=True, orient="records"
-    )
 
+def test_explore_creation(dataset_without_mapping):
+    """Test the creation of the ElasticSearch instance"""
```
dcfidalgo, line 16: To be more specific, we create an index in elasticsearch
```suggestion
    """Test the creation of an index in elasticsearch"""
```

tests/text/test_pipeline_explore.py
```diff
@@ -1,38 +1,36 @@
...
 
-def test_explore_creation(data_source_without_mapping):
     pl = Pipeline.from_config(
         {"name": "test", "head": {"type": "TextClassification", "labels": ["a"]}}
     )
     explore_id = explore.create(
-        pl, data_source_without_mapping, explore_id="mock", show_explore=False
+        pl, dataset_without_mapping, explore_id="mock", show_explore=False
```
dcfidalgo, line 22:
```suggestion
        pl, dataset, explore_id="mock", show_explore=False
```

tests/text/test_pipeline_explore.py
```diff
@@ -1,38 +1,36 @@
...
+def dataset_without_mapping() -> Dataset:
+    """Creating the dataset"""
+    data = {"text": ["This is a simple test"], "label": ["a"]}
+    return Dataset.from_dict(data)
 
-    return DataSource(
-        source=str(data_file), flatten=False, lines=True, orient="records"
-    )
 
+def test_explore_creation(dataset_without_mapping):
```
dcfidalgo, line 15:
```suggestion
def test_explore_creation(dataset):
```

tests/text/test_pipeline_explore.py
```diff
@@ -1,38 +1,36 @@
...
     )
     elastic = Elasticsearch()
     explore_meta = elastic.get(index=".biome", id=explore_id)["_source"]
     assert pl.config.as_dict() == explore_meta["pipeline_config"]
 
 
-def test_explore_with_no_mapping_in_ds(monkeypatch, data_source_without_mapping):
-    def _explore(explore_id, pipeline, data_source, options, es_dao):
-        return data_source.to_mapped_dataframe().compute()
+def test_explore_with_no_mapping_in_ds(monkeypatch, dataset_without_mapping):
```
dcfidalgo, line 29: I think we can simply remove this entire test, since we do not have datasets with our without mappings. Maybe we could move the last assert `assert len(explore_id) > 0` to the test above.

src/biome/text/explore.py
```diff
@@ -424,11 +430,18 @@ def add_predictions(batch, columns):
         num_proc=options.num_proc,
     )
 
+    # Quick fix for in-memory data that are not backed up by a file
+    # TODO: Finde a better solution
```
dcfidalgo, line 434: Typo from my side ...
```suggestion
    # TODO: Find a better solution
```

dcfidalgo (commented):
Thanks a lot! Just a few minor in line comments, then we are good to go.

tests/text/test_pipeline_explore.py
ignacioct: Okay, I will move that assert and recommit that change

tests/text/test_pipeline_explore.py
ignacioct: much clearer

ignacioct:
The first with the new assert at the end passed just fine

dcfidalgo (approved):
Great! thanks a lot!

PR #436, opened 2020-11-06
Fix/test data

ignacioct:
In this PR the test related to dataset importing (csv, json, parquet and excel) are tested.


tests/text/data/test_csv_source.py
```diff
@@ -1,17 +1,26 @@
...
-        )
-        df = ds.to_dataframe().compute()
-        self.assertIn("name", df.columns)
+@pytest.fixture
+def dataset() -> Dataset:
+    """Creation of dataset"""
+
+    file_path = os.path.join(FILES_PATH, "trailing_coma_in_headers.csv")
+
+    # TODO: erase download mode once the release which includes not caching the datasets is out
```
dcfidalgo, line 17: I think this TODO can be erased.
```suggestion
```

tests/text/data/test_json_source.py
```diff
@@ -1,34 +1,59 @@
...
-    def test_flatten_json(self):
-        file_path = os.path.join(FILES_PATH, "to-be-flattened.jsonl")
-        ds = DataSource(format="json", flatten=True, source=file_path)
-        df = ds.to_dataframe().compute()
+def test_flatten_json(dataset_flatten_source):
+    """Assert that flatten operation divides correctly"""
+    dataset_flatten_source.flatten_()
 
-        for c in ["persons.*.lastName", "persons.*.name"]:
-            self.assertIn(c, df.columns, f"Expected {c} as column name")
```
dcfidalgo, line 25: @frascuchon Do we still need this specific way of flattening `List[Dict]`s in json files (see `biome.text.data.helpers.flatten_dataframe`)? Do you remember the use case for this? If needed, we could add such a functionality directly to our `Dataset` class.

dcfidalgo (commented):
Great, LGTM. Just one small inline comment, and a question for @frascuchon.

tests/text/data/test_json_source.py
frascuchon: The use case was transform datasources to parquet easily. Parquet format cannot accept automatically data object columns.

I supose we can tackle this as a `map` operation in Dataset. So, I'm ok to drop it

dcfidalgo: :+1:

PR #437, opened 2020-11-06
Feat/precommit hook

ignacioct:
Here is a little idea I discussed with @dcfidalgo and we found it interesting to optimice the way in which we use code formatters and make the commits. It is based on [pre-commit](https://pre-commit.com/), a python package that allows the introduction of small scripts applied before each commit. It can be configured in the .yaml file attached to this PR. 

To test it, I added hooks for:

- Three predefined hooks of the pre-coomit package: checking the config file integrity, EOF fixer and trailing whitespaces fixer.
- Black hook, as if we typed `black ourscript.py` in the terminal.
- [Reorder python imports](https://github.com/asottile/reorder_python_imports), another pip package capable of reordering the imports of all script in a logical way. 

I work in VSCode, and push using the GUI functionalities built within it. In my POV, once I commit I script, I get a warning that not all changes have been committed, and I can recommit the file with all those changes applied. I expect this works in a similar way via terminal or Pycharm. 

In order to introduce this functionality to our workflow, we could make the dev version of biome to require pre-commit package (and the reorder if we may), or the interested one could include it in their personal repositories and add it to .git/info/exclude (a .gitignore that does not get uploaded to the repository).

Tell me your thoughts and opinions 😃


dcfidalgo:
Thanks a lot @ignacioct I will definitely check it out and see how it integrates with my workflow in pycharm. I will let you know when i got some experience with it.

dcfidalgo (approved):
After testing and playing around with it the last days, i think this would be a nice feature to have. It ensures a uniform code style/format (black and imports, the one-line imports lead to smaller/more specific diffs in the PRs) without having to worry about it.

@frascuchon To activate the hooks you just have to run `make dev`. Whenever you try to commit something the hooks will run first, and if they change the code, the commit is aborted and you can revise the changes. If happy, you can commit again.

ignacioct:
isort seems much more convenient to me!

dcfidalgo:
I did not include these changes in any of my PR, but directly modified this one.
I reopen it so we can merge it maybe tomorrow.

ignacioct:
Oh, I thought it was already merged. My bad, sorry.

PR #441, opened 2020-11-11
feat(dataset): create dataset from elasticsearch index

frascuchon:
This PR allows create `Dataset` instances from elasticsearch queried indices.


src/biome/text/dataset.py
```diff
@@ -134,6 +133,39 @@ def from_dict(cls, *args, **kwargs) -> "Dataset":
...
+        index
+        query
+
+        Returns
+        -------
+        A new Dataset created from scanned query records in elasticsearch index
+
+        """
+
+        from elasticsearch.helpers import scan
```
dcfidalgo, line 155: Any reason why we import this locally?

dcfidalgo (approved):
Nice!! LGTM, just a small inline question.

src/biome/text/dataset.py
frascuchon: No, bad refactoring. Change applied

PR #442, opened 2020-11-12
No objects in instance meta

dcfidalgo:
This PR resolves a bug encountered by Dani and reported on slack (see https://colab.research.google.com/drive/1GHvOdjC4cesmWuZJgU3b_lGlOsT5m14Z?usp=sharing)

With the new `Dataset` class we serialize the instances via pickle in order to cache them **and** to allow for lazy loading. In the `TokenClassification` head we were putting `MetadataField`s with spacy `Doc` objects into the instances, and pickling/unpickling those instances took quite some time plus the cached instancese were >15GB big coming from a training set of ~1MB ... 

I refactored a bit the `TokenClassification` to avoid passing on a spacy Doc via the `Instance` itself.
The down side of the approach presented here is that for predictions of untokenized text, we create two times the same spacy Doc object:
- 1. For tokenizing the input text when creating the instance
- 2. For obtaining the offsets when decoding the model output
Not sure if this can be avoided.

I also removed the automatic computation of a default value for the `num_proc` argument in `Dataset.to_instances()`, it was too sketchy ... i could expose it in the `Pipeline.train()` method, but at the moment you can already use it like this for example:
```python
pipeline.train(training=train_ds.to_instances(pipeline, num_proc=8))
```

### DIFF
--- a/src/biome/text/dataset.py
+++ b/src/biome/text/dataset.py
@@ -269,7 +269,7 @@ class Dataset:
         return self.dataset.num_rows
 
     def to_instances(
-        self, pipeline: "Pipeline", lazy=True, num_proc: Optional[int] = None
+        self, pipeline: "Pipeline", lazy=True, num_proc: int = 1
     ) -> InstancesDataset:
         """Convert input to instances for the pipeline
 
@@ -280,7 +280,7 @@ class Dataset:
         lazy
             If true, instances are lazily read from disk, otherwise they are kept in memory.
         num_proc
-            Number of processes to be spawn. If None, we try to figure out a decent default.
+            Number of processes to be spawn.
         """
         self._LOGGER.info("Creating instances ...")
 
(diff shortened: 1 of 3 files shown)


frascuchon (approved):
Good ! 

We must disclaim somewhere about what user include as allennlp Instances, and make recommendations using lightweight instance data.

dcfidalgo:
Today i found another case where we have an issue pickling the instances ... when using the transformers feature ...

The instances contain the indexers, and when using the transformers feature, pickling the corresponding indexer is very very expensive and also leads to absurdly heavy cache files (tens of GB). So i did some bench marking.

**Times for reading x instances**

| creation method | transformer feature | word feature | 
| --- | --- | --- |
| cached on disk | 5.54 s | 23.4 ms |
| created on-the-fly | 80.6 ms | 426 ms |
| cached in-memory | 2.37 us | 2.39 us |

You can see that unpickling cached instances on disk that contain a transformer feature takes ages and it is much faster (~70 times) creating the instances on-the-fly. However, using only the word feature for example, unpickling the cached instances is faster than creating them on-the-fly (~20 times). Of course, the fastest way is to store the instances in memory. I also compared these numbers using the `DataSource` class and figured that we actually persisted the dask DataFrame with the instances in memory, so it was comparable to the *cached in-memory* numbers above.

In summary, i think by default we should store the instances in memory and create them on-the-fly when we choose to lazily load them. The vast majority of our datasets so far can be kept in memory, and i think it is easy to notice when this is not the case and switch on lazy loading (`Pipeline.train(lazy=True)`). @frascuchon @dvsrepo what do you think? I will open another PR where i present this new approach.

Persisting the instances on disk would be very convenient to be able to reuse them in different executions, but i do not see an easy way to efficiently achieve this ... Regarding the transformers indexer, the problem is actually that when we create the instances in memory they share the same indexer instance, but when we pickle/unpickle them, every instance has its own indexer instance which is very expensive.

---------

Regarding this PR, i am still in favor of merging it, since it will make the instances more light weight when storing them in memory.

PR #443, opened 2020-11-13
Improve instance creation

dcfidalgo:
This PR is a follow-up PR of #442 

I found an intermediate solution for caching the instances. Instead of caching each Instance by itself, we can pickle the whole list of instances. Pickle is smart enough to deal with references to class instances, so the indexers are not a problem anymore.

The default behavior of `Dataset.to_instances` is now:
- Create and keep instances in memory + cache the whole list of instances on disk.
- Subsequent calls to `Dataset.to_instance` will reuse those cached instances and load them into memory. You can turn off this behavior by specifying `use_cache=False`.
- If `lazy=True`, lazily load data from disk and create instances on-the-fly

For now we loose the ability to multiprocess the instance creation (since we do not use the `Dataset.map` method anymore), i will try to implement it again using directly the `multiprocessing` library, let's see if i manage.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,8 @@ if __name__ == "__main__":
             "mlflow~=1.9.0",
             "elasticsearch>=6.8.0,<7.5.0",
             "ray[tune]~=1.0.0",
-            "datasets~=1.1.2"
+            "datasets~=1.1.2",
+            "tqdm>=4.49.0"
         ],
         extras_require={
             "testing": [
--- a/src/biome/text/dataset.py
+++ b/src/biome/text/dataset.py
@@ -3,17 +3,29 @@ import functools
 import inspect
 import logging
 import pickle
-from typing import Union, Dict, Iterable, List, Any, Tuple, Optional, TYPE_CHECKING
+from pathlib import Path
+from typing import (
+    Union,
+    Dict,
+    Iterable,
+    List,
+    Any,
+    Tuple,
+    TYPE_CHECKING,
+    Optional,
+    Callable,
+)
 
 import datasets
-from datasets.fingerprint import Hasher
+from allennlp import __version__ as allennlp__version__
 from allennlp.data import AllennlpDataset, AllennlpLazyDataset, Instance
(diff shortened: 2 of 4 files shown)


frascuchon (approved):
Nice work!! 

Just a thoughts sharing. The more we work with allennlp, the more I think we should make effort in totally control in training process, in order to mitigate frictions with library design (vocab, instances, indexers vs embedders,...)

A "nice to have" for me would be generate datasets  ready for `forward` pass, I mean, generating directly tensors dataset, instead of intermediate `Instance` datasets.

In this sense, could be easy/efficient serialize and deserialize tensors instead complex object structures.

To achieve this, we must disassociate from current trainer, and a quite code lines must be vomited, but I would like to center next efforts, step by step, to this approach.

What do you think @dvsrepo @dcfidalgo ???

dvsrepo:
> Nice work!!
> 
> Just a thoughts sharing. The more we work with allennlp, the more I think we should make effort in totally control in training process, in order to mitigate frictions with library design (vocab, instances, indexers vs embedders,...)
> 
> A "nice to have" for me would be generate datasets ready for `forward` pass, I mean, generating directly tensors dataset, instead of intermediate `Instance` datasets.
> 
> In this sense, could be easy/efficient serialize and deserialize tensors instead complex object structures, like `Instance`'s
> 
> To achieve this, we must disassociate from current trainer, and a quite code lines must be vomited, so I would like to center next efforts, step by step, to this approach.
> 
> What do you think @dvsrepo @dcfidalgo ???

Fully agree, we need to define what needs to be done and how to go there step by step.

This is also relevant for things like using jit, and onxx to make our training and inference more efficient and fully leveraging ray tune for things like population based training.

dcfidalgo:
> A "nice to have" for me would be generate datasets ready for forward pass, I mean, generating directly tensors dataset, instead of intermediate Instance datasets.

Yes, this would be super easy to serialize. A potential issue that comes right to my head is the tight link to the vocabulary of this type of dataset ...

Anyways, we should get together some day and discuss this in more detail.

PR #444, opened 2020-11-16
Adapt CLI to dataset

dcfidalgo:
Adapts our CLI to the new `Dataset` class.

I introduced a helper function to make Elasticsearch data available to the CLI. Basically you provide the `Dataset.from_elasticsearch` arguments via a yaml file.

Honestly, i think we could remove the `train` and `explore` command and just keep the `serve` command ...


dvsrepo:
> Adapts our CLI to the new `Dataset` class.
> 
> I introduced a helper function to make Elasticsearch data available to the CLI. Basically you provide the `Dataset.from_elasticsearch` arguments via a yaml file.
> 
> Honestly, i think we could remove the `train` and `explore` command and just keep the `serve` command ...

I agree wrt `explore`, we could definitely remove it.

Not so sure about `train` and even `evaluate` (which we could support too), I think a simple version of those two commands can be useful for automation and things like integration pipelines.

dcfidalgo:
Oks, i removed the `explore` command and added a note in the evaluate issue #406 (that we want to expose it to the CLI).

Issue #404, opened 2020-10-16
Remove the `DataSource` class and replace it with the new `Dataset` class


dcfidalgo:
TODO: With our new `Dataset` class #398 the old `DataSource` class can go away and with it the dask dependency :)

To get back all of our `read_*` methods you can always go through pandas:
```python
import pandas as pd
df = pd.read_parquet(...)
ds = Dataset.from_pandas(df)
```

The only feature that i do not know how to recover, is reading directly from a S3 storage. For now we have to download the data first to the local machine before creating the `Dataset`.

PR #445, opened 2020-11-16
Remove DataSource from explore

dcfidalgo:
This is the first in a series of PRs completely remove the `DataSource` class in our library (#404).

@ignacioct and I will take care of this providing alternately small PRs (changing at most 1 file) to make the review process easier and faster.


frascuchon (approved):
👍

PR #448, opened 2020-11-17
Fix: Remove DataSource from pipeline.py

ignacioct:
This PR should remove completely DataSource form pipeline.py


src/biome/text/pipeline.py
```diff
@@ -179,18 +188,16 @@ def find_lr(
...
             )
 
-        if isinstance(training_data, DataSource):
-            training_data = self.create_dataset(training_data)
-        elif isinstance(training_data, Dataset):
-            training_data: AllennlpLazyDataset = training_data.to_instances(pipeline=self)
+        if isinstance(training_data, Dataset):
+            training_data: AllennlpLazyDataset = training_data.to_instances(
+                pipeline=self
+            )
```
dcfidalgo, line 200: The default value of `lazy` has changed, so we have to set it to True here.
Nothing to do with the removal of `DataSource`, but let's include it in this PR as well, since i just saw it.
```suggestion
            training_data: AllennlpLazyDataset = training_data.to_instances(
                pipeline=self, 
                lazy=True,
            )
```

src/biome/text/pipeline.py
```diff
@@ -4,42 +4,54 @@
...
+from typing import Optional
+from typing import Type
+from typing import Union
 
 import numpy
 import torch
 from allennlp.commands.find_learning_rate import search_learning_rate
 from allennlp.common import Params
-from allennlp.data import AllennlpDataset, AllennlpLazyDataset, Instance, Vocabulary
+from allennlp.data import AllennlpDataset
```
dcfidalgo, line 33: Do we still need this import?

dcfidalgo (approved):
Seems it passed the CI :)
I comitted a suggestion that had nothing to do with the removal of the `DataSource`, but i included it anyways in this PR.
 Just a small comment about one import line, the rest looks good to me! Thanks a lot!!

src/biome/text/pipeline.py
ignacioct: No, I think it is useless. I have erased it.

PR #449, opened 2020-11-17
Fix and adapt second tutorial to Dataset

dcfidalgo:
This fixes the second tutorial and adapts it to the new Dataset


dvsrepo:
Just a quick comment: in the last part is it still true that input texts
must be pretokenized? Also if we output span offsets I would show that
output as well.

Can go in a separate PR

dcfidalgo:
True, i will rephrase it a bit. Now you can also pass in a string as well. But if you trained your model with pretokenized tokens, you should make sure that the pipeline uses the same tokenizer as in your preprocessing of the training data.

Actually, when providing pretokenized text, the predict should return tags and entities with token spans (where the start and end key denote the tokens of the entity), but it does not work... i will have a look at this tomorrow.

PR #450, opened 2020-11-17
Fix: DataSource erased from config.py

ignacioct:
Super quick change, it should not be anymore DataSource in the package.


src/biome/text/configuration.py
```diff
@@ -523,8 +539,8 @@ class VocabularyConfiguration:
 
     Parameters
     ----------
-    sources: 
-        List of DataSource or InstancesDataset objects to be used for data creation
+    sources:
+        List of InstancesDataset objects to be used for data creation
```
dcfidalgo, line 543:
```suggestion
        List of datasets from which to create the vocabulary
```

dcfidalgo (approved):
LGTM, thanks a lot!

PR #456, opened 2020-11-20
Make output format of `Classification` and `TokenClassification` consistent

dcfidalgo:
This PR tries to go towards the output format defined [in our backlog](https://docs.google.com/spreadsheets/d/16NnFWaP4M60B2cv_p4Ic9LCV9sJ74ohSwtm8RWhPxqY/edit#gid=0).

For the `Classification` output, this means simplifying a lot the `predict` output, leaving basically only a `labels` key.
In the backlog this key carries a dict, i switched over to a list of tuples `(label[str], prob[float])` to make sure the order is preserved from highest prob to lowest prob. 

This allows us also to keep a consistent output with the `TokenClassification` prediction output by adding the viterbi score, so the output keys of a `Pipeline.predict` in the two cases would be: 
- ~~**Classification**: `labels: List[Tuple[str, float]]`~~
- ~~**TokenClassification**: `tags: List[Tuple[List[str], float]]`, `entities: List[Tuple[List[Dict], float]]`~~

**See comment below for updated format**

Optional output keys would be: `tokens` and `attributes` as mentioned in the backlog.

@frascuchon @dvsrepo I will mark this PR as Draft, since i think we need to discuss these changes first, especially with respect to the UI.


dcfidalgo:
@frascuchon @dvsrepo Argh, on a second thought i would rather move the scores/probabilities to its own key, this avoids duplication in the `tags` and `entities`, and resolves an issue with different data types in a sequence that `Dataset`/pyarrow has. So:

- **Classification**: `labels: List[str]`, `probabilities: List[float]`
- **TokenClassification**: `tags: List[List[str]]`, `entities: List[List[Dict]]`, `scores: List[float]`

 The mapping can easily done by order.

dcfidalgo:
@frascuchon I introduced a backward compatibility function in the `explore.py` module, but i did not manage to test it with the UI, it always gives me a Segmentation fault when trying to `launch_ui` or `show`. Maybe we could have a look at this together tomorrow?

dcfidalgo:
Just checked that the UI in the first tutorial works with these changes.
@frascuchon So from my side this should be good to go.

frascuchon (approved):

Issue #438, opened 2020-11-10
[BUG] pip install from git (master) not working on Colab tutorials


dvsrepo:
# Describe the bug

`pip install` fails.

## To Reproduce
Go to and run install cell:  https://colab.research.google.com/github/recognai/biome-text/blob/master/docs/docs/documentation/tutorials/1-Training_a_text_classifier.ipynb

It might be related to recent setup.py changes.

## Expected behavior

Most important is: can we make pip install from source work on Collab with recent changes?

Most immediate solutions, either:

- Review all tutorials and modify install git url pointing to the version/commit where they were functioning.
- At least for the first, it works with the 1.0 release on PyPi: `pip install biome-text`

This is related to a broader issue, already discussed, we need to anchor tutorials either to a specific release (we actually should create releases more often) or using git commit urls to make sure they are installed with the right version.

dcfidalgo:
Just merged PR #439 that should fix the tutorial. The issue was that we now require a very new pip version for the installation to take advantage of the new dependency resolver. Not sure why it did not print the corresponding error message that we have  in our `setup.py` ... In the soon to come version 20.3 the new resolver is supposed to become the standard, so maybe colab will update the default version once it is released.

Yes, i agree tutorials should be linked to releases! For now I will try to add git commit urls as an intermediate solution.

dvsrepo:
Cool!!

PR #457, opened 2020-11-23
Introducing the transformers tutorial

dcfidalgo:
This PR adds 4th tutorial about the *transformers* feature to our documentation.
The base of this tutorial was created in collaboration with @ignacioct , who also published this [very nice report](https://wandb.ai/ignacioct/biome/reports/Exploring-Ways-to-use-Pretrained-Transformers-in-biome-text--VmlldzoyNzk2MTM) on the WandB gallery as a direct result. You can find the tutorial here: https://github.com/recognai/biome-text/blob/feat/transformers_tutorial/docs/docs/documentation/tutorials/4-Using_Transformers_in_biome_text.ipynb

I am still not 100% happy, i think some minor things can be improved (for example, using `batches_per_epoch` instead of using a subset of the data for the HPO, or using some advanced search algorithm, since parallelizing the HPO in Google Colab is not possible due to the resources). There is also a TODO left: evaluation on a test set, i will add this once we have the new evaluate method. 
But In the near future i want to go over all the tutorials anyways, unifying them a bit and trying to improve things. So for now, i think this is a good first version.

### DIFF
--- a/docs/docs/documentation/tutorials/4-Using_Transformers_in_biome_text.ipynb
+++ b/docs/docs/documentation/tutorials/4-Using_Transformers_in_biome_text.ipynb
@@ -43,7 +43,8 @@
    "outputs": [],
    "source": [
     "!pip install -U pip\n",
-    "!pip install --use-feature=2020-resolver git+https://github.com/recognai/biome-text.git"
+    "!pip install git+https://github.com/recognai/biome-text.git\n",
+    "exit(0)  # Force restart of the runtime"
    ]
   },
   {
@@ -52,8 +53,6 @@
     "id": "NLS2EK3HCfT8"
    },
    "source": [
-    "Ignore warnings and don't forget to restart your runtime afterwards *(Runtime -> Restart runtime)* .\n",
-    "\n",
     "*If* you want to log your runs with [WandB](https://wandb.ai/home), don't forget to install its client and log in."
    ]
   },
@@ -128,7 +127,8 @@
     "from biome.text import Dataset, Pipeline\n",
     "from biome.text.configuration import VocabularyConfiguration, TrainerConfiguration\n",
     "from biome.text.hpo import TuneExperiment\n",
-    "from ray import tune"
+    "from ray import tune\n",
+    "import os"
    ]
   },
   {
@@ -380,7 +380,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": null,
    "metadata": {
     "id": "26jk87DqvWxr"
    },
@@ -873,7 +873,7 @@
    },
    "outputs": [],
    "source": [
-    "analysis = tune.run(\n",
+    "analysis_frozen = tune.run(\n",
     "    tune_exp,\n",
     "    scheduler=tune.schedulers.ASHAScheduler(), \n",
     "    metric=\"validation_accuracy\",\n",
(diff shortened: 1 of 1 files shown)


docs/docs/documentation/tutorials/4-Using_Transformers_in_biome_text.ipynb
```diff
@@ -0,0 +1,1196 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "id": "TYh_I8imVosu"
+   },
+   "outputs": [],
+   "source": [
+    "!pip install -U pip\n",
+    "!pip install --use-feature=2020-resolver git+https://github.com/recognai/biome-text.git"
```
ignacioct, line 46: This install can be normal again I think. We did it to introduce Datasets if I don't remember bad, and that is integrated into biome now.

dvsrepo:
Great work! Congrats @ignacioct and @dcfidalgo.

I skimmed through the tutorial and the report and looks good to go on my side. I agree we'll come back to the tutorials to improve things but this already adds a lot of value.

Yes, it would be nice to tackle the `pipeline.evaluate()` bit before merging.

docs/docs/documentation/tutorials/4-Using_Transformers_in_biome_text.ipynb
```diff
@@ -0,0 +1,1196 @@
...
+    "\n",
+    "We also need to specify the maximum number of input tokens `max_length` supported by the pretrained transformer. If you are sure that your input data does not exceed this limit, you can skip this parameter.\n",
+    "\n",
+    "With BERT like models, like RoBERTa, a special [CLS] token is added as first token to each input. It is pretrained to effectively represent the entire input and can be used as pooler in the head component. Many BERT like models pass this token through a non-linear tanh activation layer that is part of the pretraining. If you want to use these pretrained weights you have to use the `bert_pooler` together with the corresponding `pretrained_model`. We will fine-tune these weights as well (setting `require_grad` to `True`) and add a little dropout.\n",
```
ignacioct, line 370: Super small kind of typo: I would write "with BERT-like models such as RoBERTa", to avoid the two 'like' so close. The new context lines are cool!

ignacioct (approved):
A brilliant work! I just remembered the preprocessing notebook, maybe it's not worth it to include it since the notebook is crystal clear and in that notebook we relied on Pandas, but remember that we can upload it to Kaggle or even to biome or Collab and link it.

docs/docs/documentation/tutorials/4-Using_Transformers_in_biome_text.ipynb
dcfidalgo: :+1: done

docs/docs/documentation/tutorials/4-Using_Transformers_in_biome_text.ipynb
dcfidalgo: When we added the new `Dataset` we were installing from a branch, here i install from master, so should be fine. But as Dani mentioned in #438 we should link the tutorials to a specific release or at least commit. working on this.

ignacioct: Then it is perfect!

dcfidalgo:
Ok, added the evaluation section and did some minor improvements. I will quickly merge this before it gets lost in a dusty never ending story ... :) Hopefully we get the building of the docs working still today.

PR #458, opened 2020-11-23
Introducing the mutable pipeline

dcfidalgo:
This PR switches to a "mutable" pipeline, that is training the pipeline modifies directly its weights instead the ones of a copy of the pipeline. It is a follow-up on the discussion here #410  .

I did some minor bench marking about memory consumption, and especially for big pretrained transformers this approach saves some memory depending on the size of the transformer (it was ~400 MB in case of the distilroberta_base). 

Since a mutable pipeline means that `_BlankPipeline` can be a `_PreTrainedPipeline` and `_PipelineCopy` is not needed anymore, i switched to only one `Pipeline` class. And one remark about `Pipeline.create_vocabulary`: i would like to get rid of this and make the vocab creation for a training as automatic as possible. If one wants some fancy vocab creation (most of the times you don't and when using pretrained transformers you can't) you can pass on a VocabConfiguration directly to the train method. I have some draft ready, will present this soon.

### DIFF
--- a/src/biome/text/_model.py
+++ b/src/biome/text/_model.py
@@ -2,7 +2,6 @@ import inspect
 import json
 import logging
 import os
-import pathlib
 import pickle
 import warnings
 from functools import lru_cache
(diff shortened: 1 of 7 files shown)


dcfidalgo:
@frascuchon Done, moved the `_model_path` to the `PipelineModel.file_path`. Looks cleaner this way!

frascuchon (approved):
YEAH!

dvsrepo:
Just a comment, do not forget to update tutorials if needed.

PR #461, opened 2020-11-25
fix(explore): do not stringfy predictions

frascuchon:
This PR should fixes explore UI problems


dcfidalgo (approved):
Not sure why it fixes the issue, but it works :)
Thanks!!!

PR #462, opened 2020-11-26
Add a `Pipeline.evaluate` method

dcfidalgo:
This PR adds a `evaluate` method to our `Pipeline` reusing basically AllenNLP's evaluate function.

I am not sure anymore if we should use an evaluate method in our `explore` command, therefore i refrained from a more complex custom implementation.

For our UI we need several things that an "evaluation" does not necessarily provide:
- loss per instance (basically means batch_size = 1)
- attributes (from our `explain`)

On the other side, the metrics we get out of an evaluation are also not really interesting for the UI, since we calculate the metrics there on-the-fly applying the respective filters.

Additionally, now we use a "prediction cache" in the explore method that is not really compatible with an evaluate method.. 

@frascuchon Maybe we need another technical discussion about the explore functionality. For now, if we want to have the loss per instance, we can simply use the `predict` instead of the `predict_batch` method inside the explore, which includes a loss key when providing the label as well.

### DIFF
--- a/src/biome/text/cli/cli.py
+++ b/src/biome/text/cli/cli.py
@@ -4,10 +4,11 @@ import sys
 import click
 from click import Group
 
+from .evaluate import evaluate
 from .serve import serve
 from .train import train
 
-SUPPORTED_COMMANDS = [serve, train]
+SUPPORTED_COMMANDS = [train, evaluate, serve]
 
 
 def main():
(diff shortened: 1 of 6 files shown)


dvsrepo (approved):
I'm ok with a simple evaluate method for now, and we can discuss how we go about the explore method.

Just one comment, we want the static metrics on the explore side, on the fly metrics do not scale very well when we add new models. But I agree we need to write a spec , discuss the explore feat

PR #464, opened 2020-11-27
Refactor/revise dependencies

dcfidalgo:
This PR revises a bit the dependencies:
- pyarrow not needed anymore, comes with *datasets*
- ujson not used in our library, and makes trouble in windows
- an up to date version of gevent solves a RunTimeError for greenlet and also solves the Segmentation Fault when launching the UI via `explore.show` or `biome.text.ui.launch_ui`

It also includes a small import cleanup.


frascuchon (approved):
👍

PR #465, opened 2020-11-30
feat(ds): select es dataset features

frascuchon:
This PR allows select which field will be loaded from elasticsearch documents.


dcfidalgo (approved):
Nice!

PR #467, opened 2020-12-03
Automatic vocab creation

dcfidalgo:
This PR removes the necessity of creating a vocab manually before the training. 

So in the most common use case you just have to create the pipeline, maybe a trainer config and then call the train method:
```python
pipeline = Pipeline.from_config(...)
trainer_config = TrainerConfiguration(...)
pipeline.train(...)
```
If you want more control over the vocab creation, you can pass in a vocab config:
```python
pipeline.train(..., vocab_config=VocabularyConfiguration(...))
```
The default value of vocab_config is basically `VocabularyConfiguration(datasets=[training_dataset])`.
If you set `vocab_config=None` the pipeline's vocabulary will be kept untouched, but i am not 100% sure about its use cases.

Now we basically always use the `extend_vocabulary` / `extend_vocab` method to extend our vocabulary either from an empty vocab, or from a populated vocab in case of a pretrained pipeline. I think i found a bug when we were extending the vocab before: we extended the vocab by first creating the new vocab and then adding the source vocab, which destroyed the right mapping of the embedding matrix when we extended it. 

One of the trickier part was to include the case where we have a pretrained word vector file. But i managed to keep the information of the pretrained_file in the `model.tar.gz` and if the file still exists when loading the model, it will be used when extending the vocab. If it does not exist it will simply print out a warning. Actually this warning is also printed out when we did not specify any pretrained_file, but this is a bug in allennlp, i will open an issue on there side.

As a side note, if you do not want the training to modify the vocab or weights of your pipeline you can now do:
```python
pipeline.copy().train(...)
```

@frascuchon @dvsrepo Sorry for this monster PR, most of the changes are actually tests and doc strings, but maybe we can have a quick call and go over it together in case you have further questions.

### DIFF
--- a/src/biome/text/_helpers.py
+++ b/src/biome/text/_helpers.py
@@ -2,21 +2,30 @@ import copy
 import json
 import logging
 import os
-from typing import Any, Dict, List, Optional, Tuple, cast
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
+from typing import cast
 
 import uvicorn
 from allennlp.common import Params
-from allennlp.common.util import prepare_environment, sanitize
+from allennlp.common.util import prepare_environment
+from allennlp.common.util import sanitize
 from allennlp.data import PyTorchDataLoader
 from allennlp.data.samplers import BucketBatchSampler
 from allennlp.models import archive_model
 from allennlp.models.archival import CONFIG_NAME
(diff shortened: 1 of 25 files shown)


dvsrepo:
Very nice work @dcfidalgo!

I will not go through the details, but the looking roughly at the changes they make total sense.

All good on my side.

PR #471, opened 2020-12-11
Github Actions to autodeploy website on PR and master pushes

ignacioct:
@dcfidalgo and I have made the migration from TravisCI to Github Actions. Due to the necesity of installing biome and pydoc, we have make just one job, in which we launch the package check and we `make build`. In future PR we will try to subdivide it, caching biome package so we don't need to install it twice. 

Here´s a graphical summary of the job steps. 
Captura de pantalla 2020-12-11 a las 20 38 12

We have also changed some configurations to build the site correctly, and make a correct import of the fonts (which gave us some problems).


dcfidalgo:
Great, thanks a lot!

PR #475, opened 2020-12-15
Improve CI

dcfidalgo:
This PR improves a bit our CI by applying caches. I also switched to using our conda environment file to test our development setup (see the new [contributing](https://www.recogn.ai/biome-text/documentation/community/contributing.html#setting-up-for-development) section of our docs).

I realized that we cannot avoid a biome.text installation when building the docs, so i kept the two things together in one job (tests and docs). I also tried to color the pytest output in the github action console, but for some reason this does not work ... no idea why.


dcfidalgo:
I will merge this PR, since i want to open another one touching the CI.

PR #477, opened 2020-12-16
feat(ds): select es document fields

frascuchon:
This PR allow select elastic document fields (`id`,  `_type` and `_index`) as selectable fields for `Dataset.from_elasticsearch` dataset features


dcfidalgo (approved):
lgbt!

PR #478, opened 2020-12-17
Include deploying a release in the CI

dcfidalgo:
This PR includes deploying a release in the CI. Here is the new guide on how to make a release that gets automatically published on PyPi.

**How to make a Release**
Just make a [GitHub release](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/managing-releases-in-a-repository#creating-a-release). Version tags should be `v1.1.0` or for release candidates `v1.1.0rc1`. After publishing the release, the CI gets triggered and if everything goes well the release gets published on PyPi. The CI does:
- run tests & build docs
- build package
- upload to testpypi
- install from testpypi
- upload to pypi

Under the hood the versioning of our package is now managed by [`setuptools_scm`](https://github.com/pypa/setuptools_scm), kind of the official tool by PyPA to manage versioning with git projects. It basically works with the tags in the git commit history. 

On a side note, it seems that more and more packages are moving away from `setup.py` to `pyproject.toml`. This could be some future improvement.

The current CI does not include automatic uploading of the package to our PyPi repo, maybe we can talk about that next week when discussing the roadmaps for biome-text and biome-ui.


frascuchon (approved):
cool !

PR #482, opened 2020-12-24
Allow transformers in Token/RelationClassification

dcfidalgo:
After the tests we did with @ignacioct, this PR enables transformers for the Token and RelationClassification head by using the mismatched feature of AllenNLP.
If you know what you are doing and are an expert you can now also skip the configuration checks. This allows you to easily perform advanced tests, for example using word pieces in the heads mentioned above.

```python
from biome.text.configuration import PipelineConfiguration
PipelineConfiguration._SKIP_CHECKS = True
```


ignacioct:
Looks good to me :)

PR #489, opened 2021-01-07
style versioning component

leiyre:
adapt versioning component to biome UI


dcfidalgo (approved):
Looks awesome, thanks a lot!!

I fixed a trailing whitespace issue and moved the axios dependency to the right place.
I also included a fix to the security alert we are getting in GitHub due to the `package.json` in the `ui` folder.

PR #493, opened 2021-01-13
UI fix fonts path

leiyre:


dcfidalgo:
In the end i will fix it with #496 . I realized that i had to manually change the tutorials anyways, so i included the fonts.styl in the little helper script ... sorry for the unnecessary work!!

PR #498, opened 2021-01-14
Expose model parameters

dcfidalgo:
This PR exposes the model parameters of the Pipeline. This allows us to easily freeze parameters before training, i give an example of how to do this in the method's doc string.

I used this method to try out the [BitFit](https://nlp.biu.ac.il/~yogo/bitfit.pdf) approach.


dvsrepo (approved):
👍

PR #500, opened 2021-01-21
Refactor/Rewrite of the `TaskOutput` class

dcfidalgo:
This PR is basically a rewrite/refactor of the `TaskOutput` class. It became larger than expected, but i think the follow up PR combining predict/explain will become smaller because of this.

 This is a rough list of all important changes:
- *task output* is now defined as output of a prediction; the output of the `forward` method is referred to as *forward output* through out the code
- `TaskOutput` is now a base class for specific output classes depending on the task; all output classes are *dataclasses*.
- these classes are located in `biome.text.modules.heads.task_output`
- the former `decode` method is replaced by a `make_task_output` method that works on an instance level, **not** batch level (this was recommended before by @frascuchon)
- the `forward` methods return a dict -> less friction with AllenNLP's API
- some "cosmetics" in the base `Classification` head

advantages:
- the `TaskOutput` classes clearly show what predictions of each task look like
- simplified code structure + naming is more explicit -> less complexity

disadvantage:
- There should be a tiny speed penalty for certain batch predictions. In the `ClassificationHead`s for example, we used to compute the probabilities in a batch, now we compute them per instance. I don't expect this penalty to be noticeable, but wanted to keep track of it.

### DIFF
--- a/src/biome/text/_model.py
+++ b/src/biome/text/_model.py
@@ -172,30 +172,8 @@ class PipelineModel(allennlp.models.Model):
         self._update_head_related_attributes()
 
     def forward(self, *args, **kwargs) -> Dict[str, torch.Tensor]:
-        """The main forward method. Wraps the head forward method and converts the head output into a dictionary"""
-        head_output: TaskOutput = self._head.forward(*args, **kwargs)
-        # we don't want to break AllenNLP API: TaskOutput -> as_dict()
-        return head_output.as_dict()
-
-    def make_output_human_readable(
-        self, output_dict: Dict[str, torch.Tensor]
-    ) -> Dict[str, torch.Tensor]:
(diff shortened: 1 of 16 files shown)


frascuchon (approved):
Super cool !! !Great work. 

Just one comment. Maybe `TaskOutput` is quite general naming and could lead to misunderstandings. What about `TaskPrediction` instead? It reforce the idea of task prediction results, and separate even more the ambiguity with `forward` method output...

dcfidalgo:
> Just one comment. Maybe `TaskOutput` is quite general naming and could lead to misunderstandings. What about `TaskPrediction` instead? It reforce the idea of task prediction results, and separate even more the ambiguity with `forward` method output...

Yes, this naming is much better, like it a lot. Commit is coming ...

PR #504, opened 2021-02-01
Fix: Batch predictions always return a list

dcfidalgo:
With this PR a batch prediction (`Pipeline.predict(batch=...)`) always returns a list, **also** with length == 1.


frascuchon (approved):
great!

PR #513, opened 2021-02-03
Batch predictions can contain None 

dcfidalgo:
With this PR the `Pipeline.predict` output can contain None. This is the case if for the provided input we cannot generate an instance, or the forward output cannot be converted into a `TaskPrediction` (less likely).

I added checks to the `TaskHead.featurize`s, that looks for empty TextFields, and if found the returned instance will be None. This indirectly catches empty string inputs, but be aware of tokenizers that add special tokens ... maybe i could add an explicit check for empty strings in the provided input.

Anyways, if you already know that for certain inputs the model will fail, i think it is more efficient to avoid sending this input to the model in the first place. @frascuchon let me know if you want me to add an explicit check for empty inputs in the heads.

### DIFF
--- a/src/biome/text/_model.py
+++ b/src/biome/text/_model.py
@@ -85,6 +85,7 @@ class PipelineModel(allennlp.models.Model):
     """
 
     PREDICTION_FILE_NAME = "predictions.json"
+    _LOGGER = logging.getLogger(__name__)
 
     def __init__(self, name: str, head: TaskHead):
         super().__init__(vocab=head.backbone.vocab)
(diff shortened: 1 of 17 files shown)


dcfidalgo:
@frascuchon I was not very happy with the implementation, so i reworked it a bit. The idea now is that the instance creation basically never throws an exception but returns a `None` and logs the exceptions instead. I implemented an empty strings/tokens check in the featurizer that checks for empty input before and after the tokenization (like this we could have caught our problem with the transformer tokenizer in Cantemist earlier ...). Now you can throw as many empty strings as you wish at the model without breaking it, but i still would recommend filtering them out before hand out of efficiency reasons.

The `Pipeline.predict` now only throws an `PredictionError` if all predictions are `None`. But it will always log the failing predictions as exceptions.

dcfidalgo:
@frascuchon New approach :)

Now the `TaskHead.featurize` returns either an Instance or throws an `FeaturizeError`. Catching this error is the responsibility of the calling methods. Now, both predictions and training, catch this error, skip the corresponding instance and emit a logging warning. Before, the training did basically the same (skipping Nones instead of catching the error), so i left it like that. I think it is more practical this way, not to screw up an entire training run, if only one lazy loaded instance gave problems.

Going to merge this now, i need this for the custom profner head this week.

PR #522, opened 2021-02-12
Docs: Add tip about creating an Analysis object from a directory

### LINKED ISSUE #521: [Tutorials] Add tip about readingTune analysis from experiment dir
I have to look up this every time and it's somehow not very accessible in tune docs:

```python
from ray.tune.analysis.experiment_analysis import Analysis
results = Analysis(experiment_dir='base_hpo/Baseline profner classifier', 
                   default_metric='validation_accuracy',
                   default_mode='max'
)
```
It's often the case that you have this long running HPO process and you want to revisit and analyse the results at a later stage,  the `analysis = tune.run()..` it's the ideal case for very quick HPO processes, but I guess reading from dir is much more frequent. This is a tune thing but I think it adds value to the tutorials where we have HPO stuff.

dcfidalgo:
This PR closes #521


dvsrepo (approved):
💯

PR #523, opened 2021-02-19
Make `PipelineModel` a LightningModule

dcfidalgo:
This PR makes the `PipelineModel` inherit from a `pl.LightningModule` and removes the inheritance from a `allennlp.Model`. This is the first step to allow our pipeline to be trained with the Lightning `Trainer`.

In order not to break any existing flow (that is using AllenNLP's trainer, evaluation and model serialization), i had to do two things:
- copy over a few methods from the `allennlp.Model`, mainly `extend_embedder_vocab`, `forward_on_instances` and `load`. I already customized them a bit, but they can surely be optimized or even merged with some of our existing methods, especially because now we have complete control over the whole prediction flow. I will leave this for future PRs.
- customize the method to load a pipeline from a `model.tar.gz`.

In a follow-up PR i will introduce the first version of our "Lightning" trainer.

### DIFF
--- a/src/biome/text/_model.py
+++ b/src/biome/text/_model.py
@@ -3,9 +3,11 @@ import json
 import logging
 import os
 import pickle
+import re
 import warnings
 from functools import lru_cache
 from logging.handlers import RotatingFileHandler
+from os import PathLike
 from pathlib import Path
 from typing import Any
 from typing import Dict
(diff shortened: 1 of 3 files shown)


src/biome/text/_model.py
```diff
@@ -53,7 +62,7 @@ def __hash__(self):
         return pickle.dumps(self).__hash__()
 
 
-class PipelineModel(allennlp.models.Model):
+class PipelineModel(pl.LightningModule, Registrable):
```
frascuchon, line 65: Is there any reason why it can't be extended from `allennlp.Model` as well?

```python 
class PipelineModel(allennlp.models.Model, pl.LightningModule)
```

Looking for as few changes as possible for maintainability as we tackle the transition, I would highly recommend that option, if we still have direct dependencies to allennlp.

Quite of code included in this class could be dropped.

frascuchon (commented):
Nice, just a general comment/question about `allennlp.Model` inheritance that would like to discuss

src/biome/text/_model.py
dcfidalgo: This inheritance makes this PR a bit superfluous, if your prefer to tackle this change afterwards, ok with me!
Will close this PR and open the other one for reviewing and discussions.

I will keep this branch locally, to have the changes ready when we want to make the transition to AllenNLP2.0.

PR #527, opened 2021-02-22
Introduce new lightning trainer

dcfidalgo:
~This is a follow-up PR to #523 and **must only be merged after #523** and should only be reviewed after #523 is in.~ (See comment below).

This PR introduces the new `Trainer`. The usage was discussed in [this colab](https://colab.research.google.com/drive/1pjTpCGRPsizk2p31woqYikHzrPN-Ac6h?usp=sharing) last week, you can see an example in the tests:

```python
from biome.text import Pipeline, Dataset, Trainer
from biome.text.configuration import VocabularyConfiguration

pl = Pipeline.from_config(...)
train_ds, valid_ds = Dataset(...), Dataset(...)

trainer = Trainer(    
    batch_size=64,
    optimizer={"type": "adam", "lr": 0.01},
    max_epochs=5,
)

# optional
vocab_config = VocabularyConfiguration(
    datasets=[train_ds], max_vocab_size={"word": 50}
)

trainer.fit(
    pipeline=pl,
    train_dataset=train_ds,
    valid_dataset=valid_ds,
    vocab_config=vocab_config,
)

evaluation = pl.evaluate(valid_ds)
```

I created also a new `LTrainerConfiguration` class, since mixing all the possible parameters for the allennlp and lightning trainer was super messy. The idea is to step by step include all AllenNLP trainer parameters to achieve the same feature with the lightning trainer (if possible), and then make it the default `TrainerConfiguration`.

For now the `Trainer` just updates the weights of the pipeline, prints out the metrics in the progress bar and logs them to tensorboard.

Next steps would be to serialize the model after training, early stopping, logging to mlflow + wandb, use the new trainer in `TuneExperiment` by default ...


dcfidalgo:
In the end we will tackle PR #523 when we want to make the transition to AllenNLP2.0 (and drop the allennlp trainer), and will first implement a fully working Lightning Trainer.

src/biome/text/trainer.py
```diff
@@ -0,0 +1,124 @@
...
+        self._config = copy.deepcopy(config)
+
+        # remove non pl.Trainer arguments
+        trainer_kwargs = asdict(self._config)
+        for kwarg in ["num_epochs", "optimizer", "data_bucketing", "batch_size"]:
+            del trainer_kwargs[kwarg]
+
+        self.trainer = pl.Trainer(**trainer_kwargs)
+
+    def create_dataloader(
```
frascuchon, line 35: Is it worth having this method as a helper function in order to be able to use a pytorch trainer directly?

```python
trainer= pl.Trainer(....)

trainer.fit(pipeline.model, helper.create_dataloader(dataset, batch_size=32, data_bucketing=True)
....
```

src/biome/text/trainer.py
```diff
@@ -0,0 +1,124 @@
...
+            train_dataset.to_instances(pipeline, lazy=lazy)
+        )
+        valid_dataloader = (
+            self.create_dataloader(valid_dataset.to_instances(pipeline, lazy=lazy))
+            if valid_dataset is not None
+            else None
+        )
+
+        # create optimizer
+        pipeline.model.optimizer = Optimizer.from_params(
```
frascuchon, line 111: in the same line that comment above, could we move the optimizer creation to `Pipeline` or `PipelineModel` class directly ? 

```python
pipeline.model.create_optimizer(optimizer_config)
```

frascuchon (approved):
Nice !!!

Just a few comments related to use case of using a LightningTrainer directly.

src/biome/text/trainer.py
dcfidalgo: Do you think a proper method is really necessary?

Since it will be a somehow advanced pure pytorch feature, you can always do: 
```python
from torch.optim import Adam
pipeline.model.optimizer = Adam(
    pipeline.model.parameters(), 
    lr=0.001, 
)

# or with AllenNLP stuff

from allennlp.common import Params
from allennlp.training.optimizers import Optimizer
pipeline.model.optimizer = Optimizer.from_params(
    Params(
        {
            "type": "adam",
            "model_parameters": pipeline.model.named_parameters(),
            "lr": 0.001,
        }
    )
)

```

src/biome/text/trainer.py
dcfidalgo: Done

src/biome/text/trainer.py
frascuchon: Ok, I though model optimizer was a allenlp module instead of a pytorch module.

Cool !

dcfidalgo:
@frascuchon @dvsrepo I was thinking of removing the `LTrainerConfiguration` and provide all the parameters directly to the `Trainer` class. In the example above the step of defining the `trainer_config` and passing it as exclusive argument to the `Trainer` feels a bit superfluous. What do you think?

PR #535, opened 2021-03-25
class_weights

javispp:


src/biome/text/modules/heads/classification/classification.py
```diff
@@ -30,7 +30,11 @@ class ClassificationHead(TaskHead):
     _LOGGER = logging.getLogger(__name__)
 
     def __init__(
-        self, backbone: ModelBackbone, labels: List[str], multilabel: bool = False
+        self,
+        backbone: ModelBackbone,
+        labels: List[str],
+        multilabel: bool = False,
+        class_weights: List[float] = None,
```
dcfidalgo, line 37:
```suggestion
        class_weights: Optional[List[float]] = None,
```

src/biome/text/modules/heads/classification/classification.py
```diff
@@ -39,8 +43,10 @@ def __init__(
         self._multilabel = multilabel
 
         # metrics and loss
+        if class_weights != None:
```
dcfidalgo, line 46: It's more common to use `variable is not None`, since there is only one [instance of None](https://stackoverflow.com/questions/2209755/python-operation-vs-is-not)
```suggestion
        if class_weights is not None:
```

dcfidalgo (commented):
Nice, thanks a lot :+1: Just a few inline comments before we can merge this.

dcfidalgo:
See #536

PR #540, opened 2021-03-30
Feat: Introducing LightningTrainerConfiguration

dcfidalgo:
This is a bit of a mess ...
The main change is the introduction of a `LightningTrainerConfiguration` dataclass (to be renamed when the old Trainer goes away). Until now it was (more similar to pure lightning):
```python
trainer = Trainer(**trainer_kwargs)
trainer.fit(pipeline, train_data)
```
Now it is: (more similar to hf transformers):
```python
trainer_config = TrainerConfiguration(**trainer_kwargs)
trainer = Trainer(pipeline, train_data, trainer_config)
trainer.fit()
```
I saw that i needed the pipeline and data to define sane defaults in the `Trainer` init, that is why i moved the trainer_kwargs to a proper dataclass. It is one line more, but i think it is worth it in order to have better defaults.

This PR also includes a minor change in the vocab config: i moved the `include_valid_data_in_vocab` parameter to the `VocabularyConfiguration` as `include_valid_data`. Conceptually it makes much more sense there.

### DIFF
--- a/src/biome/text/configuration.py
+++ b/src/biome/text/configuration.py
@@ -1,8 +1,11 @@
 import copy
-import dataclasses
-from typing import TYPE_CHECKING
+import os
+from dataclasses import dataclass
+from dataclasses import field
+from pathlib import Path
 from typing import Any
 from typing import Dict
+from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Type
@@ -15,6 +18,11 @@ from allennlp.common.checks import ConfigurationError
 from allennlp.data import TokenIndexer
 from allennlp.data import Vocabulary
 from allennlp.modules import TextFieldEmbedder
+from pytorch_lightning import Callback
+from pytorch_lightning.accelerators import Accelerator
+from pytorch_lightning.loggers import LightningLoggerBase
+from pytorch_lightning.plugins import Plugin
+from pytorch_lightning.profiler import BaseProfiler
 
 from .features import CharFeatures
 from .features import TransformersFeatures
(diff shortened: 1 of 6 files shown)


src/biome/text/configuration.py
```diff
@@ -644,4 +654,297 @@ class PredictionConfiguration:
 
     add_tokens: bool = False
     add_attributions: bool = False
-    attributions_kwargs: Dict = dataclasses.field(default_factory=dict)
+    attributions_kwargs: Dict = field(default_factory=dict)
+
+
+@dataclass
+class LightningTrainerConfiguration:
```
dvsrepo, line 661: My main question would be how are we going to keep up with changes and additions to the `pytorch_lightning.trainer`.  Do not know the specifics but from the outside it seems they evolve really fast, but maybe is something we can cope with, what's your opinion @dcfidalgo ?

dcfidalgo: @dvsrepo Yes, they move fast ... but i think it is not super critical for us to require the latest version. Not as critical as AllenNLP, for example, since they define our pytorch version to work with, which is important for a smooth experience in Colab.

So i think it's feasible to bump up the pl version in our requirements from time to time and update those parameters. That said, i also thought about only exposing the more frequently used parameters in the `TrainerConfiguration` and put the rest in a `lightning_trainer_kwargs`. Maybe this would make this configuration class a bit handier and would make the update to new pl versions a tiny bit easier. But for this i would like to gain a bit more experience with the arguments first.

dvsrepo: Fully agree, all good on my side, ignore my comment

PR #547, opened 2021-04-15
Activate warnings

dcfidalgo:
At some point we deactivated all warnings, i think to avoid too many warnings from allennlp. Here i reactivate them again, since we are moving away from the allennlp trainer and i think they can be useful to spot some potential issues with the usage of third party libraries. When we encounter a frequent warning we should try to fix it or specifically deactivate that warning somehow.

This PR also includes some fixes that avoid warnings and their potential issues.


dvsrepo:
Cool!

PR #555, opened 2021-05-03
improved error when text or label columns are missing

### LINKED ISSUE #425: Improve error message when the "label" column is missing in the dataset
At the moment, if the dataset is missing the "label" column (or whatever column is necessary to train the model) the error message is:
```terminal
RuntimeError: The model you are trying to optimize does not contain a 'loss' key in the output of model.forward(inputs).
```
I think we should catch this failure earlier and print out a more precise error message.

javispp:
closes #425

### DIFF
--- a/src/biome/text/pipeline.py
+++ b/src/biome/text/pipeline.py
@@ -417,11 +417,28 @@ class Pipeline:
         try:
             self.__configure_training_logging(output, quiet)
 
+            # check "text" and "label" columns in datasets
+            def check_columns(dataset, name):
+                missing_columns = []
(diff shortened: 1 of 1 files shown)


dcfidalgo (changes requested):
Thanks for working on this! 
Your presented solution, however, only works with the `TextClassification` head. Ideally we want something more general.
As a side note, `Pipeline.train` is already deprecated, so better to directly work with the new `biome.text.trainer.Tainer`.

dcfidalgo:
This PR is a bit outdated, will close it and wait for a new one.

PR #556, opened 2021-05-06
Refactor/upgrade fastapi deps

frascuchon:
Upgrade fastapi and uvicorn deps for pipeline serve. `fastapi` and `uvicorn` deps are required from a minimal version.

```python
"fastapi>=0.63.0"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 default: help
 
 dev: ## install package in development mode
-	@pip install --upgrade -e .[dev,serve]
+	@pip install --upgrade -e .[dev]
 	@pre-commit install
 
 check: ## applies a code pylint with autopep8 reformating
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 default: help
 
 dev: ## install package in development mode
-	@pip install --upgrade -e .[dev]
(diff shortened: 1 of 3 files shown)


dcfidalgo (commented):
Not sure if i would introduce the new `serve` extra requirement, i think a simple `pip install biome-text` should come with the minimal serving capabilities we offer. Maybe we can relax a bit the dependencies (for example, just require a minimal version of fastapi and uvicorn, since we only use their basic stuff) in order of avoid dependencies issues with rubrix.

If we want to keep `[serve]`, then we should add some references in our repo README and install documentation.

frascuchon:
> Not sure if i would introduce the new `serve` extra requirement, i think a simple `pip install biome-text` should come with the minimal serving capabilities we offer. Maybe we can relax a bit the dependencies (for example, just require a minimal version of fastapi and uvicorn, since we only use their basic stuff) in order of avoid dependencies issues with rubrix.
> 
> If we want to keep `[serve]`, then we should add some references in our repo README and install documentation.

Reverted. serve deps are installed by default and setted up only with minimal version requirement.

dcfidalgo:
Perfecto!

PR #557, opened 2021-05-06
remove all explore UI references

frascuchon:
For now, tutorials and documentation just omit explore predictions sections. Wi will include them (or a link to rubrix tutorials) when ready.


dcfidalgo (commented):
Nice! Just a small comment, i think without the `explore` feature we can get rid of some dependencies (for example elasticsearch). Did you check this?

frascuchon:
> Nice! Just a small comment, i think without the `explore` feature we can get rid of some dependencies (for example elasticsearch). Did you check this?

yes, elasticsearch deps are dropped.

PR #563, opened 2021-05-24
Feat: Upgrade to allennlp2

dcfidalgo:
This PR finally upgrades our dependency to AllenNLP2.0 and removes a lot of "old trainer" stuff. This allows us to use pytorch 1.8.1 and a newer transformers version. I basically only had to add the classes `AllennlpDataset` and `AllennlpLazyDataset`. 

We loose two features: `Pipeline.find_lr` and `TrainerConfiguration.data_bucketing`. The first one can still be achieved via the `Trainer.trainer.tuner`, but it's a bit tricky and i plan on facilitate this with a `Trainer.find_lr` method. The latter one is a bit uncertain, i have to have a closer look at this, but since we are not using it right now and the upgrade is more important, i removed it for now.


dvsrepo:
Great job @dcfidalgo !

PR #565, opened 2021-05-25
Add version to biome.text dependency in the mlflow model conda env

### LINKED ISSUE #564: Mlflow models could not work with new biome-text changes
**Is your feature request related to a problem? Please describe.**
Pipelines published as mlflow models could not work if biome-text changes, since versions are not fixed.

**Describe the solution you'd like**
default conda environment should include and fix the biome-text version when pipeline is published as mlflow model.

dcfidalgo:
Closes #564


frascuchon (approved):
Nice!!!, 

Just as comment (but we won't do nothing): dev biome-text versions will create unusable conda environment, isn't it?

PR #566, opened 2021-05-31
Bump up spacy to 3.0

dcfidalgo:
Our dependency on spacy is more direct than i thought. It is difficult for us to keep supporting spacy2.0 if we want to be compatible with spacy3.0:
- `Span.string` -> `Span.text`
- `spacy.gold` -> `spacy.training`, and also the method names changed (`offsets_to_biluo_tags` ...)
- no need to call `create_pipe` first, a simple `add_pipe` is sufficient
- no more model shortcuts

I vote for bumping up the spacy version and requiring spacy3.0 as this PR does (AllenNLP's spacy dep is >=2.1.0, <3.1.0). @frascuchon what do you think?

### DIFF
--- a/src/biome/text/tokenizer.py
+++ b/src/biome/text/tokenizer.py
@@ -53,8 +53,7 @@ class Tokenizer:
         if config.segment_sentences and not self.__nlp__.has_pipe(
             self.__SPACY_SENTENCIZER__
         ):
-            sentencizer = self.__nlp__.create_pipe(self.__SPACY_SENTENCIZER__)
-            self.__nlp__.add_pipe(sentencizer)
+            self.__nlp__.add_pipe(self.__SPACY_SENTENCIZER__)
 
         if config.text_cleaning is None:
             self.text_cleaning = TextCleaning()
@@ -130,7 +129,7 @@ class Tokenizer:
         if not self.config.segment_sentences:
             return list(map(self._tokenize, texts[: self.config.max_nr_of_sentences]))
         sentences = [
-            sentence.string.strip()
+            sentence.text.strip()
             for doc in self.__nlp__.pipe(texts)
             for sentence in doc.sents
         ]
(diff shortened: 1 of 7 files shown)


frascuchon:
> Our dependency on spacy is more direct than i thought. It is difficult for us to keep supporting spacy2.0 if we want to be compatible with spacy3.0:
> 
> * `Span.string` -> `Span.text`
> * `spacy.gold` -> `spacy.training`, and also the method names changed (`offsets_to_biluo_tags` ...)
> * no need to call `create_pipe` first, a simple `add_pipe` is sufficient
> * no more model shortcuts
> 
> I vote for bumping up the spacy version and requiring spacy3.0 as this PR does (AllenNLP's spacy dep is >=2.1.0, <3.1.0). @frascuchon what do you think?

I imagine a wide part of community already using spaCy 3.0 version

src/biome/text/tokenizer.py
```diff
@@ -227,16 +225,3 @@ def _tokenize(self, text: str) -> List[Token]:
...
-    # Allennlp get_spacy_model method works only for fully named models (en_core_web_sm) but no
-    # for already linked named (en, es)
-    # This is a workaround for mitigate those kind of errors. Just loading one more time, it's ok.
-    # See https://github.com/allenai/allennlp/issues/4201
-    import spacy
-
-    try:
-        spacy.load(lang, disable=["vectors", "textcat", "tagger" "parser" "ner"])
-    except OSError:
-        spacy.cli.download(lang)
```
frascuchon, line 242: we no longer need download the model if doesn't exist?

dcfidalgo: AllenNLP does this for you with its [get_spacy_model](https://github.com/allenai/allennlp/blob/cf113d705b9054d329c67cf9bb29cbc3f191015d/allennlp/common/util.py#L258)

frascuchon (approved):

PR #579, opened 2021-06-28
refactor(inference): default fallback prediction

frascuchon:
Adding a fallback prediction instead of returning a `None` object, helps inference service integrations with tabular oriented operations, like mlflow serving (expecting a dataframe as result)

### DIFF
--- a/src/biome/text/model.py
+++ b/src/biome/text/model.py
@@ -356,7 +356,7 @@ class PipelineModel(allennlp.models.Model, pl.LightningModule):
         self,
         batch: List[Dict[str, Union[str, List[str], Dict[str, str]]]],
         prediction_config: PredictionConfiguration,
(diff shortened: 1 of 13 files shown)


dvsrepo (approved):
👍

dcfidalgo (commented):
Nice! Thanks a lot! 
Only one comment: i would feel more comfortable using the name `TaskHead.empty_prediction` for the new attribute instead of `fallback_prediction`. It is more specific and for now covers all the use cases. For me it is hard to imagine a use case in which you want to return something else than an empty prediction in case the inference fails.

@frascuchon If you are ok with this name change, i can implement it, there is also some minor stuff missing, like changing the return types and adding the new attribute to all of our heads.

frascuchon:
> empty_prediction

oks!

